PR #13, opened 2016-08-10
Fix heredoc descriptions

### LINKED ISSUE #12: Support here document syntax
As mentioned in the [Configuration Syntax](https://www.terraform.io/docs/configuration/syntax.html#_lt_lt_EOT) documentation, multiline strings can be formatted using here document syntax, but as of v0.0.2, variable descriptions using here document syntax are rendered as blank values.

*input*

```hcl
variable "foo" {
  description = <<EOD
This is foo's description.
EOD

  default = "bar"
```

*expected output*

```sh
var.foo ("bar")
This is foo's description
```

*actual output*

```sh
var.foo ("bar")
-

jbussdieker:
Fixes #12


yields:
thanks!

Issue #15, opened 2016-11-17
Terraform Docs Support of Terraform Lists


pantocrator27:
As of Terraform 0.7.0, terraform supports lists in variable declarations that does not appear to be supported in Terraform Docs.

pvdvreede:
I just ran into this problem as well where a list type that has a default as `[]` does not register as an optional variable.

metmajer:
@pantocrator27 @pvdvreede Inputs of type `list` are supported by `terraform-docs` for a while. Now, also their default values can be rendered when using the `---with-aggregate-type-defaults` option with #53.

PR #16, opened 2016-11-20
Allow top-level comments for variables, and markdown line break fixes, list placeholder

vancluever:
Hi all!

Amazing tool, been looking to use this for a while but just have had the time to dive into it now. Noticed a couple of things that seemed like low-hanging fruit to fix.

First off, I've never really ever used the `description` parameter in variables, taking a godoc-style approach to documenting both variables and outputs (ie: using top-level comments). IMO it breaks workflow to have to throw docs one place and then a different place for something that's very similar in nature. Hence, I made a little backwards-compatible change that allows for the use of comments *if* description is not specified.

Second off, it looks like if one has line breaks in descriptions, the resulting markdown is broken, as markdown tables require line breaks as `<br>` tags. So the (first) commit addresses that.

**One more**: added a placeholder for list types for now. This addresses #15 somewhat, would be nice to see defaults fully rendered for both lists and maps, but at least this will properly document if a list variable is required or not.

**And another!** Updated the header comment processing logic so it handled the extra whitespace character after the `*` if it exists. This ensures that the header data is not indented mistakenly (can break markdown in a few places). This could probably be handled a bit better, but it fixed the issues I was seeing.

Thanks again for making this, is going to help me immensely when maintaining docs for our modules!

### DIFF
--- a/doc/doc.go
+++ b/doc/doc.go
@@ -23,6 +23,8 @@ func (i *Input) Value() string {
 			return i.Default.Literal
 		case "map":
 			return "<map>"
+		case "list":
+			return "<list>"
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


yields:
thanks man!! :)

vancluever:
Of course @yields and thanks for the quick turnaround! :)

PR #19, opened 2017-02-15
Added sorting to inputs and outputs

nwalke:
Very newb Go'er here, so let me know if this is terrible.


yields:
great, thanks man!

PR #22, opened 2017-03-29
Strip # prefix from comments

jacobwgillespie:
As it stands, comments like the following actually capture the `#` as part of the comment text:

```hcl
# a comment
output "something" {}
```


yields:
thanks!

PR #23, opened 2017-03-30
Add note about installing with Homebrew

jacobwgillespie:
Just got `terraform-docs` into Homebrew here: https://github.com/Homebrew/homebrew-core/pull/11768.


yields:
oh awesome, thanks man!

PR #25, opened 2017-04-06
Add support for reading `description` tag from `output` resources.

### LINKED ISSUE #24: Description tag not support on outputs
Although it's not documented on the Terraform [`outputs`](https://www.terraform.io/docs/configuration/outputs.html) page, the `description` tag is actually supported on `output` resources. See https://github.com/hashicorp/terraform/blob/76dca009e0f048ea35e86f3737e7f63503378789/config/config.go#L160-L166

Therefore it would be useful to support reading the `description` value for `output` resources, in the same was as it's supported for `variable` resources.  

PR incoming to implement.

fatmcgav:
Fixes #24

### DIFF
--- a/doc/doc.go
+++ b/doc/doc.go
@@ -120,10 +120,10 @@ func outputs(list *ast.ObjectList) []Output {
 			items := item.Val.(*ast.ObjectType).List.Items
 			var desc string
 			switch {
-			case description(items) != "":
-				desc = description(items)
 			case item.LeadComment != nil:
 				desc = comment(item.LeadComment.List)
+			case description(items) != "":
+				desc = description(items)
 			}
 
 			ret = append(ret, Output{
--- a/doc/doc.go
+++ b/doc/doc.go
@@ -117,9 +117,13 @@ func outputs(list *ast.ObjectList) []Output {
 		if is(item, "output") {
 			name, _ := strconv.Unquote(item.Keys[1].Token.Text)
 
+			items := item.Val.(*ast.ObjectType).List.Items
 			var desc string
-			if c := item.LeadComment; c != nil {
-				desc = comment(c.List)
+			switch {
+			case description(items) != "":
+				desc = description(items)
+			case item.LeadComment != nil:
+				desc = comment(item.LeadComment.List)
 			}
 
 			ret = append(ret, Output{


doc/doc.go
```diff
@@ -117,9 +117,13 @@ func outputs(list *ast.ObjectList) []Output {
...
 			name, _ := strconv.Unquote(item.Keys[1].Token.Text)
 
+			items := item.Val.(*ast.ObjectType).List.Items
 			var desc string
-			if c := item.LeadComment; c != nil {
-				desc = comment(c.List)
+			switch {
+			case description(items) != "":
+				desc = description(items)
+			case item.LeadComment != nil:
```
yields: To preserve the previous behavior, can we prefer the `LeadComment` and fallback to `description` only if `LeadComment == nil`?

yields:
Thanks man, just a single comment

doc/doc.go
```diff
@@ -117,9 +117,13 @@ func outputs(list *ast.ObjectList) []Output {
 		if is(item, "output") {
 			name, _ := strconv.Unquote(item.Keys[1].Token.Text)
 
+			items := item.Val.(*ast.ObjectType).List.Items
```
yields: I think this might panic in some cases, can we do a check like:

```
var items []*ast.ObjectItem

if obj, ok := item.Val.(*ast.ObjectType); ok {
  items = obj.List.Items
}

```

fatmcgav: I've not seen any panics thus far in my testing, and that's including modules with no outputs...

doc/doc.go
fatmcgav: Sure.. I went this way to mirror the `inputs` logic, but can switch around...

doc/doc.go
yields: sounds good!

doc/doc.go
fatmcgav: Does it make more sense to maintain the same behaviour between `inputs` and `outputs`, and just add a note to the documentation?

yields: Yeah, it definitely makes sense to me, but i don't think we should change the behavior of terraform-docs and break some docs out there :)

fatmcgav: @yields Just pushed an additional commit that puts `item.LeadComment` first for outputs.

yields: thanks!

philsttr:
Can we get a new release made with this fix?

PR #27, opened 2017-04-06
Add support for printing the variable 'type' in Markdown.

fatmcgav:
I thought it would be really useful if it was possible to document the variable type in addition to the existing documentation. So I've added support :) 

Currently only markdown supported, but trivial to add to other outputs.


yields:
thanks!

PR #44, opened 2017-12-21
If there is a description on an output, it should be considered before the preceding comment

jocgir:
We have something like that in our terraform code:

```
#------------------------------------------------------------------------------
# Output variables
#------------------------------------------------------------------------------
output "out1" {
  description = "Description of out1"
  value       = "out1"
}

output "out2" {
  description = "Description of out2"
  value       = "out2"
}
```

If the comment is directly before our first output, the comment is considered as the description of out1 even if there is an explicit description. I think that explicit description should prevail over comment.


yields:
thanks!

PR #52, opened 2018-07-12
Have Output Parameters

jmahowald:
We've found that while using outputs is great within a given terraform "project" using outputs with remote states is unwieldy.  

To use the s3 output from another terraform project
requires you to specify a bucket name, a bucket region, a key in the bucket to point at the state, and then the output name.  

Where using something like AWS parameter store
allows you to just specify a name.   As such we currently allow [AWS SSM Parameters](https://www.terraform.io/docs/providers/aws/r/ssm_parameter.html) to be documented like outputs.

This PR allows for that, or others like [Azure's Key Vault Secrets])(https://www.terraform.io/docs/providers/azurerm/r/key_vault_secret.html) to be marked as outputs


metmajer:
Hi @jmahowald.

Thank you for your interest in this project and apologies to have you wait for an answer.

With the release of [v0.4.0](../releases/v0.4.0), the project has undergone major refactoring to facilitate future community contributions like yours. Likewise, in the absence of tests, maintaining a high-quality codebase was close to impossible. Also, this has now changed.

That said, this pull request leaves a couple of question marks. Currently, I have a hard time understanding why this would be a good addition to `terraform-docs`. It seems very use-case specific. Would you like to share your thoughts so we can come to a conclusion?

Thanks!

metmajer:
@jmahowald I'm closing this pull request due to inactivity. Feel free to reopen when you're ready. Cheers!

PR #115, opened 2019-06-24
Enhance Makefile and add editorconfig

### LINKED ISSUE #67: Integrate 'gofmt' and 'goimports' into Makefile
### Prerequisites

Put an `x` into the box that applies:

- [ ] This issue describes a bug.
- [x] This issue describes a feature request.

For more information, see the [Contributing Guidelines](https://github.com/segmentio/terraform-docs/tree/master/CONTRIBUTING.md).

### Description

Integrating `gofmt` and `goimports` into the Makefile helps contributors code formatting and package imports automated.

### LINKED ISSUE #88: terraform-docs --version output (dev build) #86
### Prerequisites

Put an `x` into the box that applies:

- [ ] This issue describes a bug.
- [x] This issue describes a feature request.

For more information, see the [Contributing Guidelines](https://github.com/segmentio/terraform-docs/tree/master/CONTRIBUTING.md).

### Description
This is a follow up to #86 . As it is not possible to reopen an issue.

Execution of `terraform-docs --version` on Linux only outputs `dev` if it was installed using the given `go get` command.

#### Actual Behavior

Output:

```
# terraform-docs --version
dev

```

Do you have long logs to share? Please go to https://ghostbin.com and insert the link here.

#### Expected Behavior

Expected output is more than just `dev`. E.g. a string including a build date. "Old" dev builds might be used to e.g. create bug reports without being able to refer to the installed version.
```
dev (build dd-mm-yyyy HH:mm)
```

### Replication Case

Tell us which steps to take to replicate your problem.

1. go get -v github.com/segmentio/terraform-docs
2. terraform-docs --version

### Version

You can get this information via `terraform-docs --version`.
dev
Whichever dev build it might be.

khos2ow:
### Prerequisites

### Description

This PR enhances the `Makefile` and its targets. List of changes:

- add `help` target and colorize the output logs
- add `verify`: Verify vendor dependencies
- add `fmt` and `checkfmt` for go files formatting
- add `build-all` which builds for all the specified OS/ARCH as well as sha256sum file
- add _release_ helper targets: `major`, `minor` and `patch` which auto upgrade version to corresponding logical next version
- add `tools` which installs any required tools for development or CI jobs
- add `-mod=vendor` and make sure the usage of Makefile targets will be universal, no matter if you're running it within GOPATH or outside of it.

Also added `.editorconfig` file to enforce unified file coding, line ending, tab/space sizes etc on all different IDEs and text-editors.

Added a `checkfmt` job on the CI pipeline as well to make sure we test code formatting on each PR as well. This means that if that job fails the formatting of go files are incorrect and can be fixed manually or by running `make fmt` from the root of the repo.

### Issues Resolved

Fixes #67 
Fixes #88 

### Checklist

#### Tests

- [ ] I have added tests to cover my changes.

#### Documentation

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

#### Code Style


metmajer:
@khos2ow thank you for your submission! As I see it, your pull request addresses many diverse things that should be split up into dedicated submissions. Obvious candidates are the `.editorconfig` (which I find really useful), a new approach to versioning and creating releases, and potentially some more. I'd also like to you to take some time to explain this approach and why we need it.

khos2ow:
To be honest I didn't want to create two PRs with only one file change in each (one for `editorconfig` and one for `.golangci.yml`), but I'll be happy to extract these to their own PRs if you see fit.

In terms of `Makefile` refactoring, those are the practices I've gathered over the while and potentially will be really useful. One of the main advantages is that you are not bound to be in `GOPATH` for the make targets to work (this is my own setup)

The release helper targets will potentially make the release process easier for the maintainers. When the time comes what needs to be done is essentially to decide which version bump is needed. Let's assume we want to go from `v0.6.0` to `v0.7.0` the only command needed is:

```
cd /path/to/local/clone
git checkout master
git pull # make sure the master is up to date
make minor push=true
```

This will bump the version, create a tag with that version and push it to remote.

And potentially the release artifacts can be generated locally with `make build-all` and manually uploaded to github release page.

khos2ow:
@metmajer do you want to proceed with this PR? or any suggestion on breaking it down into smaller PRs?

metmajer:
Yes, I do. Could you please...

1) Adapt the `.editorconfig` to reflect the current indentation scheme? Right now, we use tabs, not spaces. Also, why use a tab size of 8 in the Makefile if you otherwise suggest 4 spaces? I suggest we use 4 everywhere.
2) I am currently not convinced by the semi-automated versioning scheme. I just don't see the benefit over editing the VERSION file manually. Can we remove the versioning magic from this PR for now?

Thanks!

khos2ow:
1. That's true, actually another section should be added in `.editorconfig` for `.go` files to enforce `tab`, although `gotfmt` already takes care of this, but that should've been added.

2. The reason I specifically set tab size of 8 for Makefile was to get the correct visual alignment. Because most of the editors consider tab size of 4 instead of actual 8. Consider the following for tab size of 4 and 8 for Makefile:

Screenshot at 2019-08-08 23-32-46

Screenshot at 2019-08-08 23-35-55

3. IMO maintaining current version of binary in `VERSION` file is both redundant and error prone. Considering that we're releasing through git tags, and those are predictable and discoverable (programmatically) we don't really need to capture the same thing in a file.
And also the process of releasing a new version will be slightly harder if we do have such a file. Maintainer needs to bump the version in that file, commit and push it, and then tag the branch.
To be fair, if we choose to keep the `VERSION` file, it will still be easier to execute a make target (such as `make minor`) which does all the above rather than do them manually.

khos2ow:
- added `.go` in `.editorconfig` and set a tab size of 4 for it.
- moved long script from Makefile to their own external files and removed the tab size of 8 for Makefile.

Makefile
```diff
@@ -1,59 +1,149 @@
...
 NAME        := terraform-docs
 VENDOR      := segmentio
 DESCRIPTION := Generate docs from Terraform modules
 MAINTAINER  := Martin Etmajer <metmajer@getcloudnative.io>
 URL         := https://github.com/$(VENDOR)/$(NAME)
 LICENSE     := MIT
 
-VERSION     := $(shell cat ./VERSION)
+# Repository variables
+PACKAGE     := github.com/$(VENDOR)/$(NAME)
```
metmajer: Since we now have ˋPACKAGEˋ and ˋURLˋ has never been used, let‘s remove the latter!

Makefile
```diff
@@ -1,59 +1,149 @@
...
 
-GOBUILD     := go build -ldflags "-X main.version=$(VERSION)"
-GOPKGS      := $(shell go list ./... | grep -v /vendor)
+# Build variables
```
metmajer: Can you please sort the variables in each block where possible?

Makefile
```diff
@@ -1,59 +1,149 @@
...
-build-freebsd-amd64:
-	GOOS=freebsd GOARCH=amd64 $(GOBUILD) -o bin/freebsd-amd64/$(NAME)
+.PHONY: verify
+verify: ## Verify 'vendor' dependencies
+	@ $(MAKE) --no-print-directory log-$@
+	$(GOCMD) mod verify
 
-build-linux-amd64:
-	GOOS=linux GOARCH=amd64 $(GOBUILD) -o bin/linux-amd64/$(NAME)
+.PHONY: lint
```
metmajer: Please sort the make targets in each block.

Makefile
```diff
@@ -1,59 +1,149 @@
...
+lint: ## Run linter
```
metmajer: I am not sure these comments are too helpful... lots of redundancy, hmm?

Makefile
```diff
@@ -1,59 +1,149 @@
...
+build-all: GOOS      = linux darwin windows freebsd
+build-all: GOARCH    = amd64 arm
+build-all: clean ## Build binary for all OS/ARCH
+	@ $(MAKE) --no-print-directory log-$@
+	@ ./hack/build/build-all-osarch.sh "$(BUILD_DIR)" "$(NAME)" "$(VERSION)" "$(GOOS)" "$(GOARCH)" $(GOLDFLAGS)
+
+#####################
+## Release targets ##
+#####################
+.PHONY: release patch minor major
```
metmajer: Let‘s keep the pattern of having a .phony above each target.

Makefile
```diff
@@ -1,59 +1,149 @@
...
+	@ $(MAKE) --no-print-directory log-$@
+	curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s  -- -b $(shell go env GOPATH)/bin $(GOLANGCI_VERSION)
+	curl -sfL https://github.com/git-chglog/git-chglog/releases/download/$(GITCHGLOG_VERSION)/git-chglog_$(shell go env GOOS)_$(shell go env GOARCH) -o $(shell go env GOPATH)/bin/git-chglog && chmod +x $(shell go env GOPATH)/bin/git-chglog
+	GO111MODULE=off go get -u github.com/mitchellh/gox
+
+####################################
+## Self-Documenting Makefile Help ##
+####################################
+.PHONY: help
+help:
```
metmajer: Appears as magical. Here, and also in the log target, some documentation would really be beneficial.

.chglog/CHANGELOG.tpl.md
```diff
@@ -50,4 +50,4 @@
 [{{ .Tag.Name }}]: {{ $.Info.RepositoryURL }}/compare/{{ .Tag.Previous.Name }}...{{ .Tag.Name }}
 {{ end -}}
 {{ end -}}
-{{ end -}}
\ No newline at end of file
+{{ end -}}
```
metmajer: What‘s the change here?

metmajer:
Thanks for applying the changes, @khos2ow. Meanwhile, I have added some additional feedback. Aside, please...

1) Rename the directory „hack“ to „scripts“.
2) Should we invoke git-chglog as part of a release? Until now, I have used the Git commit message „Update Changelog.“, but it shows up in the Changelog itself. We should check, if specific (this) commit messages can be excluded in the git-chglog config. Can you look into this?

Makefile
khos2ow: `URL` is not used but it's kind of part of the metadata of the repository, the same as `LICENSE`. I prefer to keep it, and may be even not using any variables in it.

Makefile
khos2ow: +1

Makefile
khos2ow: +1

Makefile
khos2ow: +1

.chglog/CHANGELOG.tpl.md
khos2ow: This was the result of applying `insert_final_newline = true` from `.editorconfig`.

khos2ow:
> Should we invoke git-chglog as part of a release?

I really like the idea, it can be done in `scripts/release/release.sh` right after `git tag --annotate ....`, so immediately after releasing a tag we can generate and update `CHANGELOG` and push it too.

Makefile
khos2ow: +1

Makefile
```diff
@@ -3,7 +3,7 @@ NAME        := terraform-docs
 VENDOR      := segmentio
 DESCRIPTION := Generate docs from Terraform modules
 MAINTAINER  := Martin Etmajer <metmajer@getcloudnative.io>
-URL         := https://github.com/$(VENDOR)/$(NAME)
+URL         := https://github.com/segmentio/terraform-docs
```
metmajer: I don’t understand why you’re removing variable extrapolation now. This should be reverted, even if URL is currently unused.

metmajer:
I totally agree with you on handling the brew formula update in a different PR, but a new release target should include an update of the Changelog. The updated Changelog should be contained in the tag, too.

Makefile
khos2ow: Sure.

metmajer: Thanks!

khos2ow:
I just revisited the doc of git-chglog and the easiest (and cleanest) way we can do, from now on, is to prefix commit with keywords, such as `feat:`, `fix:`, `doc:`, ... and only generate the changelog for subset of them. The changelog config of the git-chglog itself looks like [this](https://github.com/git-chglog/git-chglog/blob/master/.chglog/config.yml):

```
options:
  commits:
    filters:
      Type:
        - feat
        - fix
        - perf
        - refactor
  commit_groups:
    title_maps:
      feat: Features
      fix: Bug Fixes
      perf: Performance Improvements
      refactor: Code Refactoring
  header:
    pattern: "^(\\w*)\\:\\s(.*)$"
    pattern_maps:
      - Type
      - Subject
```

So instead of doing crazy hack to exclude `Update Changlog` commit, we can follow the convention mentioned above. What do you think about this approach Martin?

metmajer:
The problem with the approach you‘ve mentioned is that this project has never followed structured commits. This is why our current git-chglog looks the way it is. If we were to introduce these groups now, old commit messages would not be shown anymore. Is there no easier approach? What if we check for and ignore that commit message in the template? Also, I am not a fan of structuring commit messages. Thoughts?

metmajer:
Also, this hack can‘t be crazier than your version number hack in the Makefile 😀

khos2ow:
Oh I see. I'm trying to see if I can make the config work to be able to generate the old commit message and generate new format from now on. If not successful manually remove the "Update Changelog" from the list.

khos2ow:
Ok I've managed to make changelog both backward and forward compatible:
- it keeps the previous changlog records intact
- it considers `<type>: <subject>` as the new commit format (the same as git-chglog's Changelog itself)

@metmajer how do you want to proceed with this?

metmajer:
Awesome! That‘s the approach we need. Will you make your changes part of this PR? If we require specific commit message prefixes, we may have to amend CONTRIBUTING.md. Hm?

khos2ow:
I think PR of its own would be better, because it will change `.chglog/*` and as you mentioned `CONTRIBUTING.md`

metmajer:
Agree that we don‘t need to change CONTRIBUTING.md. If you already have a Changelog template available, I‘d really prefer to have it in here to complete the release target. If more work will be needed, let‘s conclude the PR here and add the template in an upcoming PR very soon.

khos2ow:
This commit 280835e will solve the template, but I prefer to have it in its own PR. If you feel like this is the correct place for it just proceed with this PR, otherwise let me know to revert that and open its own PR.

khos2ow:
There's only one gotcha point, that we have to start using the new format for commit message `<type>: <subject>` **from** start of the next release, otherwise the changelog for current release (`v0.7.0`) might get corrupted.

Issue #62, opened 2018-09-23
Add support for Terraform 0.12's Rich Value Types


metmajer:
### Prerequisites

Put an `x` into the box that applies:

- [ ] This issue describes a bug.
- [x] This issue describes a feature request.

For more information, see the [Contributing Guidelines](https://github.com/segmentio/terraform-docs/tree/master/CONTRIBUTING.md).

### Description

With the upcoming v0.12, Terraform will introduce the concept of [Rich Value Types](https://www.hashicorp.com/blog/terraform-0-12-rich-value-types). These rich value types will introduce breaking changes to `terraform-docs`.

metmajer:
This is going to be a beast. Interested in a challenge @khos2ow? :-)

khos2ow:
Yes, let's do this, I can make some time in this madness of schedule I'm in!

Acen:
FYI Terraform has now been updated to `0.12`. https://www.terraform.io/upgrade-guides/0-12.html

metmajer:
Yes, we're aware.

antonbabenko:
While there is no native support for Terraform 0.12 in terraform-docs, there is an alternative solution which involves `awk` script (kudos to @osterman and @cytopia for making this).

I have updated [pre-commit-terraform](https://github.com/antonbabenko/pre-commit-terraform) to [v1.13.0](https://github.com/antonbabenko/pre-commit-terraform/releases/v1.13.0) where this is implemented.

See this [pull request to terraform-aws-vpc module](https://github.com/terraform-aws-modules/terraform-aws-vpc/pull/288) for example.

metmajer:
Thanks @antonbabenko, this helps for the time being!

metmajer:
I am all with you: the missing support for Terraform 0.12 is temporary. However, if we just make it work somehow by throwing away existing functionality, the solution may work for you, but not for many others. In this case, we're not talking about an easy transition and HashiCorp has already warned us about using Terrafom 0.12's public APIs which may change without warning. Integrating with HCL 2.0 directly would be the official way, but is far from straight forward.

Please also keep in mind that, while we do our best giving this project the ongoing support it needs, we do this in our spare time. At the moment, we are deeply involved in projects to make our living. Again, this doesn't mean that we don't hear you.

In the meantime, please see @antonbabenko's suggestion above. If he, as a maintainer of many Terraform community modules and a long-time user of `terraform-docs` can live with this temporary solution, you will hopefully too for the time being.

[...]

PR #116, opened 2019-06-29
Use Cobra CLI instead of docopt

khos2ow:
### Prerequisites

### Description

This PR replaces `docopt` with [Cobra](https://github.com/spf13/cobra) CLI library. Using Cobra will give flexibility in terms of new subcommands or flags, and grouping them together.

Commands:

- `json` to generate json formatted
- `pretty` to generate colorized pretty output
- `markdown` (`md` as alias) to generate Markdown format
  - `table` (`tbl` alias) to generate Markdown Table format
  - `document` (`doc` alias) to generate Markdown Document format
- `version` to print the version of binary
- `completion` to prepare SHELL completion
  - `bash` shell completion
  - `zsh` shell completion
- `help` to show help message

Flags:

- Global
  - `--no-sort` omit sorted rendering of inputs and outputs
  - `--sort-inputs-by-required` sort inputs by name and prints required inputs first
  - `--with-aggregate-type-defaults` print default values of aggregate types 
  - `--version` to print the version of binary
  - `-h, --help` to show help message for binary and each subcommands
- Only for `markdown` and its children commands:
  - `--no-required` omit "Required" column when generating Markdown
  - **Note that this distinction to tie a flag to specific commands or subcommands wasn't possible previously**

The migration was 99% backward compatible and without any functionality change **except** the default execution of binary shows help message and `pretty` print has been moved to its own standalone commands (Any comments or suggestion on this will be greatly appreciated)

### Issues Resolved

List any existing issues this pull request resolves.

### Checklist

#### Tests

- [ ] I have added tests to cover my changes.

#### Documentation

- [x] I have updated the documentation accordingly.

#### Code Style


metmajer:
@khos2ow `docopt` is less strict than `cobra` when it comes to options and argument positions. Hence, I expect that your change will break existing usages of `terraform-docs`. Since we do not have a use-case that requires the use of `cobra`, I see that the risks outweigh the value of this pull request.

I value your support for this project a lot. But note that the Contribution Guidelines ask for contributions to be discussed with the maintainers before filing a submission. This is meant to avoid work that is not in the best interest of the project. There are a lot of open issues that still seek for contributors and I'd be happy to support you if you want to submit a pull request for one of these.

khos2ow:
Actually that's not correct, flag positioning in cobra is really flexible and can be provided before or after commands or subcommands, or even after the `[PATH...]`

Two other main benefits that Cobra gives but `docopt` doesn't, are:

- auto completion for different SHELLs
- tie flag to specific command (e.g. `--no-reguired` flag which only belongs to `markdown` command and its subsequent subcommands

And the bonus point (which might be useful eventually) is command aliases, e.g.

```
terraform-docs md tbl ./my-module
```

is completely valid.

metmajer:
@khos2ow Thanks for pointing this out. Shall we agree to reopen this after #62? Please head there for a recent update and call-to-action.

khos2ow:
Sure, I like to be able to include this in the current release, and personally like it to be before #62 (if we're going to consider this PR at all) to prevent resolving merge conflict in the future. But I'm ok either way.

metmajer:
@khos2ow this is indeed a valuable and very well implemented addition to the code base and will help us better address some of the upcoming issues. Thanks for your patience!

Issue #183, opened 2020-01-28
`terraform-docs/examples$ terraform output --json` throws errors caused by v.11 syntax


gshel:
### Prerequisites

Put an `x` into the box that applies:

- [x] This issue describes a bug.
- [ ] This issue describes a feature request.

### Description

Terraform v0.12.19 cli commands fail inside of the terraform-docs/examples directory, throwing  errors caused by v.11 syntax.

Relates to issue #165. 

#### Actual Behavior
`terraform-docs/examples$ terraform output --json`
```
Warning: Quoted references are deprecated

  on main.tf line 27, in data "aws_caller_identity" "current":
  27:   provider = "aws"

In this context, references are expected literally rather than in quotes.
Terraform 0.11 and earlier required quotes, but quoted references are now
deprecated and will be removed in a future version of Terraform. Remove the
quotes surrounding this reference to silence this warning.

(and one more similar warning elsewhere)


Warning: Quoted type constraints are deprecated

  on variables.tf line 9, in variable "string-2":
   9:   type        = "string"

Terraform 0.11 and earlier required type constraints to be given in quotes,
but that form is now deprecated and will be removed in a future version of
Terraform. To silence this warning, remove the quotes around "string".

(and 4 more similar warnings elsewhere)


Error: Invalid output name

  on outputs.tf line 16, in output "output-0.12":
  16: output "output-0.12" {

A name must start with a letter or underscore and may contain only letters,
digits, underscores, and dashes.
```
There's loads of others which require updating to v.12 syntax as well.

#### Expected Behavior
`terraform-docs/examples$ terraform output --json`
```
{
  "output-0-12": {
    "sensitive": false,
    "type": "string",
    "value": ""
  },
  "output-1": {
    "sensitive": false,
    "type": "string",
    "value": "1"
  },
  "output-2": {
    "sensitive": false,
    "type": "string",
    "value": "2"
  },
  "unquoted": {
    "sensitive": false,
    "type": "string",
    "value": ""
  }
}
```

### Replication Case

Tell us which steps to take to replicate your problem.

1. Install Terraform v.12
2. Clone Terraform-docs repo
3. `cd terraform-docs/examples`
4. `terraform output --json`

### Version

Terraform v0.12.19
terraform-docs version v0.8.0-alpha linux/amd64 BuildDate: 2020-01-28T14:36:47-05:00

gshel:
[...]

PR #191, opened 2020-02-06
Extract and render output values from Terraform

### LINKED ISSUE #165: Inject `terraform output` into `terraform-docs` as an additional 'Value' field
### Prerequisites

Put an `x` into the box that applies:

- [ ] This issue describes a bug.
- [x] This issue describes a feature request.

### Description
I'd like to have `terraform-docs` add an optional field for Outputs which would specify the current value of an output variable as it's found in state via `terraform output`.

My use-case:
The outputs of my Terraform project get imported into the Terraform projects of my users as variables. Ideally, my documentation would inform users what value to expect when they use them.

Let me know if any of that could use further clarification or if you have any feedback or suggestions.

Cheers!

gshel:
### Prerequisites

- [X] This pull request adds a feature.
- [ ] This pull request enhances existing functionality.

### Description

This PR allows users to pass the global flag '--output-values PATH' to inject a project's outputs' values into the chosen format.

1. `cd intoYourTF/project`
2. `terraform outputs --json > output_values.json`
3. `terraform-docs markdown table --output-values output_values.json`

### Issues Resolved
#165 

### Checklist

#### Tests

#### Documentation

- [TBD] I have updated the documentation accordingly.

#### Code Style


gshel:
Some things I can think off of the top of my head that I would like to hear your input on:

- If an output value is `""` should that display as `n/a`?

- If an output value is `""` should json/yaml omit the empty field?

- I looked over the documentation, noticing that there are not currently sections detailing flags. Since you marked my issue with the `documentation` label, I wanted to confirm that you'd like it added into the readme.

- I have some clunky if/else statements that I'm wondering if there are alternative, elegant solutions for (e.g. `internal/pkg/tfconf/module.go:142`)

internal/pkg/print/pretty/pretty.go
```diff
@@ -52,8 +52,7 @@ const (
 			{{- printf "\n" -}}
 			{{- range . }}
 				{{ printf "output.%s" .Name | colorize "\033[36m" }}
-				{{ tostring .Description | trimSuffix "\n" | default "n/a" | colorize "\033[90m" }}
-				{{ if $.Settings.OutputValues }}{{ .Value | trimSuffix "\n" | default "n/a" | colorize "\033[90m" }}{{ end}}
+				{{ tostring .Description | trimSuffix "\n" | default "n/a" | colorize "\033[90m" }}{{ if $.Settings.OutputValues }}{{- printf "\n" -}}{{ .Value | trimSuffix "\n" | default "n/a" | colorize "\033[90m" }}{{ end }}
```
gshel: So ugly, but needed to keep the `pretty` output the same.

cmd/root.go
```diff
@@ -51,6 +54,8 @@ func init() {
 	rootCmd.PersistentFlags().BoolVar(new(bool), "no-sort", false, "do no sort items")
 	rootCmd.PersistentFlags().BoolVar(&settings.SortByRequired, "sort-by-required", false, "sort items by name and print required ones first")
 
+	rootCmd.PersistentFlags().StringVar(&outputValuesPath, "output-values", "", "inject output values from `terraform output --json` into outputs")
```
khos2ow: Ok now I see your point about this new flag, I then suggest to split that into two flags:

- `--output-values bool` : to indicate if output values need to get extract or not
- `--output-values-from string` : path of the json file to read the output from (defaults to `terraform-output.json`

Then we would be able to internally use terraform binary itself to extract json of output and parse it rather than forcing one extra step to the user. Although sometime a user might choose to do that extra step manually for whatever reason which will be fine too.

cmd/root.go
```diff
@@ -74,7 +79,7 @@ func Execute() error {
 }
 
 func doPrint(paths []string, fn func(*tfconf.Module) (string, error)) {
-	module, err := tfconf.CreateModule(paths[0])
+	module, err := tfconf.CreateModule(paths[0], outputValuesPath)
```
khos2ow: Since we're starting to add more variables to `CreateModule` function I'd suggest we create a struct of its own, and have that as the only argument for `CreateModule(option Option)`. something like, `module.Option` in `tfconf/options.go` which has the following:

- `Path string`
- `OutputValues bool`
- `OutputValuesFile string`

internal/pkg/print/markdown/table/table.go
```diff
@@ -64,10 +64,10 @@ const (
 		{{ if not .Module.Outputs }}
 			No output.
 		{{ else }}
-			| Name | Description |
-			|------|-------------|
+			| Name | Description | {{if $.Settings.OutputValues}}Value |{{end}}
```
khos2ow: Please move the leading space from before {{if &.Settings....` to inside the if.

And also put surrounding spaces inside `{{ ... }}`, e.g. `{{ end }}` instead of `{{end}}`.

internal/pkg/print/markdown/table/table.go
```diff
@@ -64,10 +64,10 @@ const (
...
+			|------|-------------|{{if $.Settings.OutputValues}}-------|{{end}}
```
khos2ow: Same surrounding spaces.

internal/pkg/print/markdown/table/table.go
```diff
@@ -64,10 +64,10 @@ const (
...
 			{{- range .Module.Outputs }}
-				| {{ name .Name }} | {{ tostring .Description | sanitizeTbl }} |
+				| {{ name .Name }} | {{ tostring .Description | sanitizeTbl }} | {{if $.Settings.OutputValues}}{{ .Value | sanitizeTbl }} |{{end}}
```
khos2ow: Same as above again.

internal/pkg/testutil/testing.go
```diff
@@ -16,7 +16,7 @@ func GetExpected(goldenFile string) (*tfconf.Module, string, error) {
 		return nil, "", err
 	}
 
-	module, err := tfconf.CreateModule(path)
+	module, err := tfconf.CreateModule(path, "")
```
khos2ow: See my note above for `module.Option`.

internal/pkg/print/pretty/pretty.go
```diff
@@ -52,7 +52,7 @@ const (
 			{{- printf "\n" -}}
 			{{- range . }}
 				{{ printf "output.%s" .Name | colorize "\033[36m" }}
-				{{ tostring .Description | trimSuffix "\n" | default "n/a" | colorize "\033[90m" }}
+				{{ tostring .Description | trimSuffix "\n" | default "n/a" | colorize "\033[90m" }}{{ if $.Settings.OutputValues }}{{- printf "\n" -}}{{ .Value | trimSuffix "\n" | default "n/a" | colorize "\033[90m" }}{{ end }}
```
khos2ow: To keep it consistent with input values, I think this would work better. this will show the output value in `(...)` in front of the output name and without any color, the same as inputs.

```suggestion
				{{- if $.Settings.OutputValues -}}
					{{- printf " " -}}
					({{ .Value | trimSuffix "\n" | default "n/a" }})
				{{- end }}
				{{ tostring .Description | trimSuffix "\n" | default "n/a" | colorize "\033[90m" }}
```

internal/pkg/tfconf/output.go
```diff
@@ -4,6 +4,7 @@ package tfconf
 type Output struct {
 	Name        string   `json:"name" yaml:"name"`
 	Description String   `json:"description" yaml:"description"`
+	Value       string   `json:"value,omitempty" yaml:"value,omitempty"`
```
khos2ow: This will have some limitation as it will only render the value as `string` and also omit them if they are empty, which means the schema of json will be different based on the selected flags. You don't need to worry about this, we will have to circle back to it when the above comments are done.

Also I still need to understand how `terraform output --json` exactly works, I've briefly tested something and got different results based on the version of terraform binary I used.

khos2ow:
> * If an output value is `""` should that display as `n/a`?
> * If an output value is `""` should json/yaml omit the empty field?

We definitely should not drop the empty field, I don't have an answer for it at the moment, check out https://github.com/segmentio/terraform-docs/pull/191#discussion_r376729640

> * I looked over the documentation, noticing that there are not currently sections detailing flags. Since you marked my issue with the `documentation` label, I wanted to confirm that you'd like it added into the readme.

There are some documentation available [here](https://github.com/segmentio/terraform-docs/tree/master/docs/formats). I'm currently working on an enhancement to auto-generate that from whatever exists in [`cmd/`](https://github.com/segmentio/terraform-docs/tree/master/cmd). I put the `documentation` label to not forget about it.

> * I have some clunky if/else statements that I'm wondering if there are alternative, elegant solutions for (e.g. `internal/pkg/tfconf/module.go:142`)

That's true :) we need to address that after/with this https://github.com/segmentio/terraform-docs/pull/191#discussion_r376729640

internal/pkg/print/pretty/pretty.go
khos2ow: Please take a look at https://github.com/segmentio/terraform-docs/pull/191#discussion_r376729470

internal/pkg/tfconf/output.go
gshel: When you run the latest terraform, are you getting errors like I mentioned here?  #183 
 https://github.com/segmentio/terraform-docs/issues/183

internal/pkg/print/pretty/pretty.go
gshel: - [X] Make line 55 prettier

internal/pkg/print/markdown/table/table.go
gshel: - [X] Fix spacing

internal/pkg/print/markdown/table/table.go
gshel: - [X] Fix Spacing

internal/pkg/print/markdown/table/table.go
gshel: - [X] Fix Spacing

cmd/root.go
khos2ow: I've just pushed #193, you can take a look and let me know. We can proceed with that one first and then you rebase, or we can proceed with this PR (disregarding my comment above) and then I'll rebase the other one. Feel free to let me know which one you prefer.

internal/pkg/testutil/testing.go
khos2ow: This also can be disregarded considering #193

internal/pkg/tfconf/output.go
khos2ow: No it wasn't like #183 I was getting different `type` and `value`. Some of them were plain `string` and `"some value"`, some other were `["list", "string"]` and `["val1", "val2"]` and more. I haven't had enough time to try to understand a pattern.

cmd/root.go
gshel: I gave this some thought. What do you think of this alternative?

User passes `--output-values PATH` to specify a file from which to take output values.
OR
User passes `--output-values` to use the default PATH of `terraform-output.json` which is output by `terraform output --json > terraform-output.json`

The way I currently have set up has the default PATH set to an empty string, but I could easily change that to `terraform-output.json` and say that if the variable `outputValuesPath` is `""`, execute `terraform output --json > terraform-output.json`.

Two flags to do essentially the same thing sounds a bit :stuck_out_tongue_closed_eyes: to me, but it's entirely subjective.

cmd/root.go
gshel: Looks like yours is a lot more complete than what I've done for this portion. I can rebase after you merge :smile_cat:

khos2ow: Cool, so I'm gonna merge that and please rebase here and add required properties to the new `Options`.

internal/pkg/tmpl/template.go
```diff
@@ -160,6 +160,24 @@ func builtinFuncs(settings *print.Settings) map[string]interface{} {
 		"sanitizeTbl": func(s string) string {
 			return markdown.SanitizeItemForTable(s, settings)
 		},
```
gshel: @khos2ow How do you feel about the below changes? This is how I was able to keep the type in the generated print out. If there are any cases I have forgotten, let me know!

internal/pkg/tfconf/module.go
```diff
@@ -124,24 +126,35 @@ func CreateModule(options *Options) (*Module, error) {
 		}
 	}
```
gshel: I will change the env variable name if this ends up being an accepted solution - suggestions welcome!

internal/pkg/tfconf/output.go
```diff
@@ -1,11 +1,29 @@
 package tfconf
 
+import (
+	"encoding/json"
+)
+
 // Output represents a Terraform output.
 type Output struct {
```
gshel: By changing the Value attribute's type to `interface{}`, we are able to include typing in the generated printouts.

cmd/root.go
khos2ow: To be fair the two flags I mentioned don't do the same thing, `--output-values bool` is the indicator that we want to have outputs processed, and `--output-values-from string` is to specify from which JSON file we should read them, otherwise internally using `terraform` binary to attempt to fetch it within the execution of terraform-docs command.

cmd/root.go
```diff
@@ -88,7 +90,8 @@ func FormatterCmds() []*cobra.Command {
 
 func doPrint(path string, fn func(*tfconf.Module) (string, error)) {
 	options := &tfconf.Options{
-		Path: path,
+		Path:         path,
+		OutputValues: settings.OutputValues,
```
khos2ow: You actually defined `OutputValuesPath string` in `tfconf.Options` but don't pass it from here. I'd suggest to create a corresponding flag for it and pass it down.

internal/pkg/print/markdown/table/table.go
```diff
@@ -64,10 +64,10 @@ const (
 		{{ if not .Module.Outputs }}
 			No output.
 		{{ else }}
-			| Name | Description |
-			|------|-------------|
+			| Name | Description |{{ if $.Settings.OutputValues }} Value |{{ end }}
+			|------|-------------|{{ if $.Settings.OutputValues }}-------|{{ end }}
 			{{- range .Module.Outputs }}
-				| {{ name .Name }} | {{ tostring .Description | sanitizeTbl }} |
+				| {{ name .Name }} | {{ tostring .Description | sanitizeTbl }} |{{ if $.Settings.OutputValues }} {{ .Value | sanitizeInterface | sanitizeTbl }} | {{ end }}
```
khos2ow: There's an extra space right before the last `| {{ end }}` which should be removed (that's a markdown lint warning)

internal/pkg/print/settings.go
```diff
@@ -14,6 +14,10 @@ type Settings struct {
 	// scope: Markdown
 	MarkdownIndent int
 
+	// OutputValues inject output values from `terraform output --json` into outputs (default: false)
+	// scopt: Global
+	OutputValues bool
```
khos2ow: I'm thinking we don't need this flag here since it's already defined in `tfconf.Options` and this is in fact an _option_ to load the module with not a _setting_ to render and format result.

internal/pkg/tfconf/module.go
```diff
@@ -122,20 +126,57 @@ func CreateModule(options *Options) (*Module, error) {
...
 
+	options.OutputValuesPath = os.Getenv("FOO")
+	// if OutputValuesPath is empty, set the default and generate the file
+	if options.OutputValues && options.OutputValuesPath == "" {
+		err := os.Setenv("FOO", "terraform-outputs.json")
```
khos2ow: I'm completely against the use of environment variable. Actually to rephrase if environment variable should be used they should be immediately parsed right from the get go (a.k.a inside `cmd/root.go`) and the parsed values should be passed down.

But in this particular case I cannot think of a reason where we should need to use this. Having a flag is extremely visible and self-explanatory but environment variables are not and error prone. Also autocomplete works for the flags and not the environment variables.

internal/pkg/tmpl/template.go
khos2ow: I'm gonna take a closer look, but by the look of it it seems great. Although I'm also working on another PR to streamline the use of "default values" and not actually depend on `interface{}` as a literal. I'm gonna have to combine these change of yours and mine together to see if we would need both or either. I'm gonna circle back on this soon.

internal/pkg/tfconf/module.go
```diff
@@ -122,20 +126,57 @@ func CreateModule(options *Options) (*Module, error) {
...
+		if err != nil {
+			log.Fatal(err)
+		}
+		_, err = exec.Command("cd", options.Path, "&&", "terraform", "output", "--json", ">>", options.OutputValuesPath).Output()
```
khos2ow: This looks the way I had thought, I'm gonna take a close look shortly.

internal/pkg/tfconf/module.go
khos2ow: Please look at https://github.com/segmentio/terraform-docs/pull/191#discussion_r379701094

internal/pkg/print/markdown/table/table.go
gshel: Okay, fixing that now. Btw sorry if my PR has become a bit of a clusterf. I'm not super accustomed to squashing/resolving merge conflicts :stuck_out_tongue: Will make changes for this shortly!

khos2ow: No, it's completely find, not squashing commits and PR of this size is actually a better thing, makes with code review. And I'll squash merge the PR anyways, so no worries!

internal/pkg/tfconf/module.go
```diff
@@ -122,20 +126,57 @@ func CreateModule(options *Options) (*Module, error) {
...
+	var terraformOutputs map[string]*TerraformOutput
+	if options.OutputValues {
+		byteValue, err := ioutil.ReadFile(options.OutputValuesPath)
+		if err != nil {
+			return nil, fmt.Errorf("caught error while reading the terraform outputs file at %s: %v", options.OutputValuesPath, err)
+		}
+		if err := json.Unmarshal(byteValue, &terraformOutputs); err != nil {
+			return nil, err
+		}
+	}
```
khos2ow: Please extract these into a function, like `loadOutputValues`. The flow of code will be cleaner and it will be easier to test individual function and helps for mocking (if need be)

internal/pkg/tfconf/module.go
```diff
@@ -122,20 +126,57 @@ func CreateModule(options *Options) (*Module, error) {
...
+			outputs = append(outputs, &Output{
+				Name:        output.Name,
+				Description: String(outputDescription),
+				Position: Position{
+					Filename: output.Pos.Filename,
+					Line:     output.Pos.Line,
+				},
+			})
+		}
 	}
```
khos2ow: This looks cleaner and more readable to me:

```
	for _, o := range mod.Outputs {
		outputDescription := o.Description
		if outputDescription == "" {
			outputDescription = readComment(o.Pos.Filename, o.Pos.Line-1)
		}
		output := &Output{
			Name:        o.Name,
			Description: String(outputDescription),
			Position: Position{
				Filename: o.Pos.Filename,
				Line:     o.Pos.Line,
			},
		}
		if options.OutputValues {
			output.Value = terraformOutputs[output.Name].Value
		}
		outputs = append(outputs, output)
	}
```

internal/pkg/tfconf/module.go
gshel: Removed the env variable, let me know what you think~

internal/pkg/tfconf/module.go
gshel: Refactored with your original suggestions!

internal/pkg/testutil/testing.go
```diff
@@ -1,34 +1,38 @@
...
 	"os"
 	"path/filepath"
 	"runtime"
-
-	"github.com/segmentio/terraform-docs/internal/pkg/tfconf"
 )
 
-// GetExpected returns 'example' Module and expected Golden file content
-func GetExpected(goldenFile string) (*tfconf.Module, string, error) {
-	path, err := getExampleFolder()
```
gshel: I wasn't expecting the `GetExpected` function to also generate the module, so I broke it into two separate functions:

1. GetExpected: Read the Golden File and assign to expected
2. GetModule: Accept options and generate the module used for the `actual` var

internal/pkg/tfconf/module.go
gshel: Thanks!

internal/pkg/tfconf/module.go
gshel: All done~

internal/pkg/print/settings.go
gshel: Removed from settings :heavy_check_mark:

cmd/root.go
gshel: Fixed this~

cmd/root.go
gshel: Refactored with your suggestions~

gshel:
I'm not sure how we can test `--outputValues` without including the path to my pre-generated `output_values.json`. I ran into issues running `terraform output --json` inside the examples folder because I'm using Terraform v0.12.19. Ideas?

internal/pkg/testutil/testing.go
khos2ow: Yes, I've noticed this and have already had it converted to exactly like what you did in #195! Nice work.

internal/pkg/testutil/testing.go
```diff
@@ -9,26 +9,31 @@ import (
 	"github.com/segmentio/terraform-docs/internal/pkg/tfconf"
 )
 
-// GetExpected returns 'example' Module and expected Golden file content
-func GetExpected(goldenFile string) (*tfconf.Module, string, error) {
-	path, err := getExampleFolder()
+// GetModule returns what is generated by the test params
+func GetModule(options *tfconf.Options) (*tfconf.Module, error) {
+	path, err := GetExampleFolder()
```
khos2ow: Although I didn't get why you made `GetExampleFolder` exported since it's only being used in `internal/pkg/testutil/testing.go`. I'd suggest to revert it back to `getExampleFolder` (effectively make it private)

internal/pkg/testutil/testing.go
```diff
@@ -9,26 +9,31 @@ import (
...
 	if err != nil {
-		return nil, "", err
+		return nil, err
 	}
+	return module, err
+}
 
+// GetExpected returns 'example' Module and expected Golden file content
+func GetExpected(goldenFile string) (string, error) {
 	expected, err := readGoldenFile(goldenFile)
```
khos2ow: Also having a separate `readGoldenFile` seems to be irrelevant and its content can be moved here. something like this:

```go
// GetExpected returns 'example' Module and expected Golden file content
func GetExpected(format, name string) (string, error) {
	path := filepath.Join(testDataPath(), format, name+".golden")
	bytes, err := ioutil.ReadFile(path)
	if err != nil {
		return "", err
	}
	return string(bytes), nil
}
```
and delete `readGoldenFile` function altogether.

internal/pkg/tfconf/output.go
khos2ow: I just opened PR #196 which addresses this (well `Input`, but this is similar). I think it's fine we keep this change of yours (i.e. using `interface{}`) as is here and I'll rebase mine later when we get this PR merged. :+1:

internal/pkg/tmpl/template.go
khos2ow: As my above comment, I think this is good for now and we can take required actions if needed in #196 after rebase.

internal/pkg/tfconf/module.go
```diff
@@ -22,8 +25,7 @@ type Module struct {
 	Providers      []*Provider `json:"providers" yaml:"providers"`
 	RequiredInputs []*Input    `json:"-" yaml:"-"`
 	OptionalInputs []*Input    `json:"-" yaml:"-"`
-
-	options *Options
+	Options        *Options    `json:"-" yaml:"-"`
```
khos2ow: Ok I see why you had to make `Options` here exported, we shouldn't do that, because `Module` will eventually published publicly and consumer shouldn't have access to it (look #195 for detail if interested). It was my bad, I originally suggested to completely remove `OutputValues bool` from `print.Settings` which we should move it back.

And to be clear we need `OutputValues bool` in both `print.Settings` and `tfconf.Options`, because I'm trying to put a boundaries between the two in #195 . I mean `tfconf` package must not have dependency to `print` package.

internal/pkg/print/pretty/pretty.go
```diff
@@ -52,7 +52,11 @@ const (
 			{{- printf "\n" -}}
 			{{- range . }}
 				{{ printf "output.%s" .Name | colorize "\033[36m" }}
-				{{ tostring .Description | trimSuffix "\n" | default "n/a" | colorize "\033[90m" }}
+				{{- if $.Module.Options.OutputValues -}}
```
khos2ow: Please take a look at my comment below regarding `Settings` and `Options` and not exporting `Module.Options` and use `$.Settings.OutputValues` here.

internal/pkg/tfconf/module.go
```diff
@@ -187,3 +198,24 @@ func loadProviders(requiredProviders map[string]*tfconfig.ProviderRequirement, r
...
 	return providers
 }
+
+func loadOutputValues(options *Options) (map[string]*TerraformOutput, error) {
+	if options.OutputValuesPath == "terraform-outputs.json" {
+		_, err := exec.Command("cd", options.Path, "&&", "terraform", "output", "--json", ">>", options.OutputValuesPath).Output()
+		if err != nil {
+			log.Fatal(err)
+		}
+	}
```
khos2ow: This is the only remaining items, from where I can see. I'm gonna test this locally today. But my initial thought at the moment is that we actually should not set a default value for `--output-values-from`, but let me gather my thoughts on it first. Will sync up with you shortly.

internal/pkg/print/settings.go
khos2ow: Please take a look at https://github.com/segmentio/terraform-docs/pull/191#discussion_r380339458 , this needs to get reverted.

internal/pkg/testutil/testing.go
gshel: Oops leftover from something else I had tried. Thanks for catching that.

internal/pkg/testutil/testing.go
gshel: Was wondering what the `format` var is meant to be? Previously, `GetExpected` was taking the golden file name less the extension as the arg.

khos2ow: That was copy paste issue :see_no_evil: which we will need in the PR #195 which refactors stuff. For this PR only `name` suffices.

gshel: :smile:  sounds good! I figured, but wanted to clarify. Will be pushing shortly!

internal/pkg/tfconf/module.go
```diff
@@ -122,20 +123,29 @@ func CreateModule(options *Options) (*Module, error) {
...
+				Line:     o.Pos.Line,
 			},
-		})
+		}
+		if options.OutputValues {
+			terraformOutputs, err := loadOutputValues(options)
+			if err != nil {
+				log.Fatal(err)
+			}
+			output.Value = terraformOutputs[output.Name].Value
```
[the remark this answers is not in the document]
gshel: That's so true! Great catch

internal/pkg/tfconf/output.go
khos2ow: So the behavior seems to be as:

- terraform 0.12.x: `type` of output is always `[]interface{}`
- terraform 0.11.x and prior: `type` of outputs is always `string`

In our case it doesn't make any difference at the moment because we're not showing the _type_ of output and we only need to extract it properly, which by using `Type interface{}` in `TerraformOutput` struct this will be fine.

internal/pkg/tfconf/output.go
```diff
@@ -34,3 +35,10 @@ func (a outputsSortedByPosition) Swap(i, j int) {
 func (a outputsSortedByPosition) Less(i, j int) bool {
 	return a[i].Position.Filename < a[j].Position.Filename || a[i].Position.Line < a[j].Position.Line
 }
+
+// TerraformOutput is used for unmarshalling `terraform outputs --json` into
+type TerraformOutput struct {
+	Sensitive bool        `json:"sensitive"`
+	Type      string      `json:"type"`
```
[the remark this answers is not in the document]
gshel: Perhaps at some point in the future, something cool can be done with this!

internal/pkg/testutil/testing.go
```diff
@@ -9,23 +9,30 @@ import (
...
 	path, err := getExampleFolder()
 	if err != nil {
-		return nil, "", err
-	}
-	options := &tfconf.Options{
-		Path: path,
+		return nil, err
 	}
+	options.Path = path
+	options.OutputValuesPath = path + options.OutputValuesPath
```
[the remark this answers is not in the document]
gshel: Shoot I didn't remove the trailing slash from my tests. Everything seemed to pass though. Should I update the tests to remove them?

internal/pkg/tfconf/module.go
khos2ow: I've had some problem with this to make it work, but here is the working snippet

```go
func loadOutputValues(options *Options) (map[string]*TerraformOutput, error) {
	var out []byte
	var err error

	if options.OutputValuesPath == "" {
		cmd := exec.Command("terraform", "output", "-json")
		cmd.Dir = options.Path
		if out, err = cmd.Output(); err != nil {
			return nil, fmt.Errorf("caught error while reading the terraform outputs: %v", err)
		}
	} else {
		if out, err = ioutil.ReadFile(options.OutputValuesPath); err != nil {
			return nil, fmt.Errorf("caught error while reading the terraform outputs file at %s: %v", options.OutputValuesPath, err)
		}
	}

	var terraformOutputs map[string]*TerraformOutput
	err = json.Unmarshal(out, &terraformOutputs)
	if err != nil {
		return nil, err
	}
	return terraformOutputs, err
}
```

As you can see I'm not using `options.OutputValuesPath` at all unless it was explicitly passed by the user, and as such I suggest to remove the default value for it in `cmd/root.go` as well.

gshel: I'll make that change this morning!

PR #223, opened 2020-03-24
Allow hiding the "Sensitive" column in markdown

julienduchesne:
### Prerequisites

- [x] This pull request adds a feature.

### Description

Use case: for projects where all outputs are expected to not be sensitive, it adds noise. This PR adds the `--no-sensitive` option to the markdown generation command. This options removes the sensitive column. The column will show by default, this is an opt-out

### Checklist

#### Tests

#### Documentation

- [x] I have updated the documentation accordingly.

#### Code Style

### DIFF
--- a/cmd/markdown.go
+++ b/cmd/markdown.go
@@ -37,6 +37,7 @@ var mdDocumentCmd = &cobra.Command{
 
 func init() {
 	markdownCmd.PersistentFlags().BoolVar(new(bool), "no-required", false, "do not show \"Required\" column or section")
+	markdownCmd.PersistentFlags().BoolVar(new(bool), "no-sensitive", false, "do not show \"Sensitive\" column or section")
 	markdownCmd.PersistentFlags().BoolVar(new(bool), "no-escape", false, "do not escape special characters")
 	markdownCmd.PersistentFlags().IntVar(&settings.MarkdownIndent, "indent", 2, "indention level of Markdown sections [1, 2, 3, 4, 5]")
 
--- a/internal/format/document.go
+++ b/internal/format/document.go
@@ -113,7 +113,9 @@ const (
 					{{- $sensitive := ternary .Sensitive "<sensitive>" .GetValue -}}
 					Value: {{ value $sensitive | sanitizeDoc }}
 
-					{{ if $.Settings.ShowSensitivity }}Sensitive: {{ ternary (.Sensitive) "yes" "no" }}{{ end }}
+					{{ if $.Settings.ShowSensitivity -}}
+						Sensitive: {{ ternary (.Sensitive) "yes" "no" }}
+					{{- end }}
 				{{ end }}
 			{{ end }}
 		{{ end }}
--- a/internal/format/document.go
+++ b/internal/format/document.go
@@ -113,7 +113,7 @@ const (
 					{{- $sensitive := ternary .Sensitive "<sensitive>" .GetValue -}}
 					Value: {{ value $sensitive | sanitizeDoc }}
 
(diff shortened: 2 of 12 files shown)


cmd/markdown.go
```diff
@@ -37,6 +37,7 @@ var mdDocumentCmd = &cobra.Command{
 
 func init() {
 	markdownCmd.PersistentFlags().BoolVar(new(bool), "no-required", false, "do not show \"Required\" column or section")
+	markdownCmd.PersistentFlags().BoolVar(new(bool), "no-sensitive", false, "do not show \"Sensitive\" column or section")
```
khos2ow: Why did you only scoped this flag to `markdown` commands and not globally? I mean the same noise can be seen on other output formats as well.

internal/format/document_test.go
```diff
@@ -467,3 +468,29 @@ func TestDocumentHeaderFromFile(t *testing.T) {
...
 }
+
+func TestDocumentOutputValuesNoSensitivity(t *testing.T) {
+	assert := assert.New(t)
+	settings := testutil.Settings().WithSections().With(&print.Settings{
+		OutputValues:    true,
+		ShowSensitivity: false,
+	}).Build()
+
+	expected, err := testutil.GetExpected("table", "table-OutputValuesNoSensitivity")
```
khos2ow: These should be `document`

internal/format/document_test.go
```diff
@@ -467,3 +468,29 @@ func TestDocumentHeaderFromFile(t *testing.T) {
...
+	options, err := module.NewOptions().With(&module.Options{
+		OutputValues:     true,
+		OutputValuesPath: "output_values.json",
+	})
+	assert.Nil(err)
+
+	module, err := testutil.GetModule(options)
+	assert.Nil(err)
+
+	printer := NewTable(settings)
```
khos2ow: This should be `NewDocument`

cmd/markdown.go
julienduchesne: In other formats, like YAML or JSON, it seems less troublesome to me since it's just another value in the data format. In markdown, it takes away some horizontal space. Does that make sense?

internal/format/document_test.go
julienduchesne: :door: :walking_man: oops

cmd/markdown.go
khos2ow: That's true and I completely agree about the issue in markdown format, but for the sake of consistency between formats I think it would be good to have it global. But at the moment there's no easy way to have this flag available to `tfconf.Output` for marshalling.

Let me think about it a bit, at the end of the day we can proceed with your proposed change as is and figure out a cleaner way to make this globally available for other formats.

internal/format/document.go
```diff
@@ -113,7 +113,7 @@ const (
 					{{- $sensitive := ternary .Sensitive "<sensitive>" .GetValue -}}
 					Value: {{ value $sensitive | sanitizeDoc }}
 
-					Sensitive: {{ ternary (.Sensitive) "yes" "no" }}
+					{{ if $.Settings.ShowSensitivity }}Sensitive: {{ ternary (.Sensitive) "yes" "no" }}{{ end }}
```
khos2ow: What do you think about using this instead, it's more readable.

```suggestion
					{{ if $.Settings.ShowSensitivity -}}
						Sensitive: {{ ternary (.Sensitive) "yes" "no" }}
					{{- end }}
```

julienduchesne:
Thanks for the quick responses and merge. Really appreciated :smile:

PR #224, opened 2020-03-24
--no-header should not attempt reading main.tf file

julienduchesne:
### Prerequisites

- [x] This pull request fixes a bug.
- [ ] This pull request enhances existing functionality.

### Description

Our projects do not have that file. Even with `--no-header`, terraform-docs crashes. This PR stops the tool from reading a file when `--no-header` is set.
I'm not sure if the test is in the correct location. Let me know if that works for you

### Checklist

#### Tests

#### Documentation

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

#### Code Style

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -25,6 +25,10 @@ var rootCmd = &cobra.Command{
 			return !val
 		}
 		settings.ShowHeader = oppositeBool("no-header")
+		if !settings.ShowHeader {
+			options.HeaderFromFile = ""
+		}
+
 		settings.ShowInputs = oppositeBool("no-inputs")
 		settings.ShowOutputs = oppositeBool("no-outputs")
 		settings.ShowProviders = oppositeBool("no-providers")
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -25,9 +25,7 @@ var rootCmd = &cobra.Command{
 			return !val
 		}
 		settings.ShowHeader = oppositeBool("no-header")
-		if !settings.ShowHeader {
-			options.HeaderFromFile = ""
-		}
+		options.ShowHeader = settings.ShowHeader
 
 		settings.ShowInputs = oppositeBool("no-inputs")
 		settings.ShowOutputs = oppositeBool("no-outputs")
--- a/internal/format/document_test.go
+++ b/internal/format/document_test.go
@@ -504,11 +504,12 @@ func TestDocumentEmpty(t *testing.T) {
 		ShowOutputs:   false,
 	}).Build()
 
-	expected, err := testutil.GetExpected("document", "document-Empty")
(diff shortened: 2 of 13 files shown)


cmd/root.go
```diff
@@ -25,6 +25,10 @@ var rootCmd = &cobra.Command{
 			return !val
 		}
 		settings.ShowHeader = oppositeBool("no-header")
+		if !settings.ShowHeader {
+			options.HeaderFromFile = ""
```
khos2ow: I don't like this approach, I mean we need to treat everything in `Settings` and `Options` as immutable so instead of changing the value of `HeaderFromFile` on the fly based on some condition I'd suggest to create a new property in `module.Options`, something like `ShowHeader bool` with default to `true` and just set `options.ShowHeader = settings.ShowHeader` instead of these lines.

internal/format/document_test.go
```diff
@@ -494,3 +494,27 @@ func TestDocumentOutputValuesNoSensitivity(t *testing.T) {
...
+		ShowProviders: false,
+		ShowInputs:    false,
+		ShowOutputs:   false,
+	}).Build()
+
+	expected, err := testutil.GetExpected("document", "document-Empty")
+	assert.Nil(err)
+
+	options := module.NewOptions()
+	options.HeaderFromFile = "" // An empty header file means it will be ignored. --no-header will clear the file name
```
khos2ow: The above comment applies here too.

internal/format/document_test.go
```diff
@@ -494,3 +494,27 @@ func TestDocumentOutputValuesNoSensitivity(t *testing.T) {
 	assert.Nil(err)
 	assert.Equal(expected, actual)
 }
+
+func TestDocumentEmpty(t *testing.T) {
```
khos2ow: I like this test, can you create it for all the rest of the formats too?

internal/module/module.go
```diff
@@ -39,10 +39,18 @@ func loadModule(path string) (*tfconfig.Module, error) {
...
-	header, err := loadHeader(options.Path, options.HeaderFromFile)
-	if err != nil {
-		return nil, err
+	var (
+		header string
+		err    error
+	)
+
+	if options.HeaderFromFile != "" {
+		header, err = loadHeader(options.Path, options.HeaderFromFile)
```
khos2ow: Considering a new properties in `Options`, I'd suggest to change to `loadHeader(options *Options)` and check the condition immediately inside the function instead of at the higher level here. The main reason is in that case unit testing the function is completely self-contained and will be accurate.

julienduchesne:
Thanks for reviewing. I agree with your comments. Done!

PR #377, opened 2021-02-04
Use goreleaser for release workflow and tasks

khos2ow:

Use [GoReleaser](https://goreleaser.com/) for managing releases.

I have:

Manually mimiced a release and tried it locally.


.goreleaser.yml
```diff
@@ -0,0 +1,89 @@
...
+      - linux
+      - freebsd
+      - windows
+    goarch:
+      - amd64
+      - arm
+    goarm:
+      - 7
+    ignore:
+      - goos: darwin
```
mbarrien, line 23: Note that with the release the Mac M1 MacBook Pros, Darwin ARM is now something that should be supported. Right now they do support translating Intel to ARM, but having a truly native build would be useful.

khos2ow: That's a good point. Based on your comments on brew formula, would it be enough if we build with `GOOS=darwin` and `GOARCH=arm64`?

khos2ow: Actually it seems that wouldn't be possible, https://github.com/goreleaser/goreleaser/pull/1956 attempts to fix it but it has a hard dependency on go 1.16.

Issue #387, opened 2021-02-11
the modules section is always included in the resulting file


aston-r:
Hi,

```
terraform-docs -v
terraform-docs version 0.11.0 078ac5d linux/amd64 BuildDate: 2021-02-10T22:18:25Z
```

I tried the following command:
```
terraform-docs --show-all=false --show inputs md . <-is executed succesfully but it shows `Modules + Inputs` sections
terraform-docs --hide=modules md . <- does not hide the modules section
```

It looks like only `modules` sections is affected since `terraform-docs --hide=resources md .` works and hides the resources section

PR #391, opened 2021-02-12
added tests for cli/config

HeikoNeblung:

added tests for cli/config to prevent issues like #387

I have:

successful run of make test

### DIFF
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -91,6 +91,46 @@ func TestMarkdownTableCommandExecute(t *testing.T) {
 			params: []string{"md", "table", "--hide-all", "--show", "resources", filepath.Join("testdata", "example")},
 		},
 		{
+			name:   "table-NoHeader",
+			format: "markdown",
+			params: []string{"md", "table", "--show-all", "--hide", "header", filepath.Join("testdata", "example")},
+		},
+		{
+			name:   "table-NoInputs",
+			format: "markdown",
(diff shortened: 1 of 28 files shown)


HeikoNeblung:
@khos2ow can you please KeePass this PR Open. After Dome additional research i'd like to add a test dir the cobra command.

khos2ow:
Definitely, you can convert it to draft for now until it's ready. And thanks for the effort, it looks awesome 👍

HeikoNeblung:
@khos2ow I've added some test cases for the cobra command. Do you think there are more cases needed, because these are mostly the same as the format tests.

khos2ow:
I really like the idea of adding tests for `cmd/` commands in general, but it's a bit redundant with `internal/format` tests. But then again at the same time I've been thinking about refactoring `format` tests for a while now, mainly due to high percentage of conflict they impose on any single change on parallel PRs. To be honest I don't have an idea for it yet.

That being said I'll like to keep this PR around as a base for doing some experimentation and brainstorming for enhancing the test suites. I'll try to circle back to it shortly.

khos2ow:
Please hold off any major work on testing until #410 is done.

HeikoNeblung:
@khos2ow what do you think about a small function / script which generates the golden files? I‘ve created a SetExpected function testutil not committed yet) wich helps me in this particular case.

khos2ow:
I originally thought about this, but this sounds more like of a chicken and egg problem. The script to generate golden files for test _is_ eventually the thing we want to test to make sure the intended output is generated properly from. #411 might solve most of the issues on this regard.

PR #394, opened 2021-02-15
Switch to scratch as docker base image

pujan14:

Switch to scratch as docker base image.

I have:

By running `make test` and by testing newly build docker image updated dockerfile.


khos2ow (changes requested):
This is awesome, thanks @pujan14, but please apply the same changes to `scripts/release/Dockerfile` as well.

pujan14:
Hi @khos2ow,
I have updated `scripts/release/Dockerfile` as well.

khos2ow (changes requested):
Sorry @pujan14 for the long delay. Big +1 on removing unused packages! But after second thought and playing around with your changes, I'd like to ask you to revert back to use `alpine`. The reason is that with `scratch` will effectively block any downstream usage of the tool. For example a user cannot extend it, or install additional packages inside it, or even they cannot change the entrypoint to go _inside_ the container.

Please let us know if you'd still be interested to work on this PR.

pujan14:
@khos2ow I'll update this PR with this feedback.

PR #406, opened 2021-02-23
Completely remove deprecated '--no-XXX' flags

khos2ow:

In v0.10.0 flags '--no-XXX' have been deprecated in favor of their
counterparts in the format of '--X=false' (e.g. --no-header and
--header=false). And they are completely being removed now (two releases
after original announcement)

Signed-off-by: Khosrow Moossavi <khos2ow@gmail.com>

I have:

TBA.


khos2ow:
@HeikoNeblung these flags need to be removed and I was thinking to proceed with this PR first before #391. Otherwise it might impose some redundant work that eventually will get removed. What do you think? We can go ahead with this one and then you can rebase your PR and we can get it from there.

HeikoNeblung:
@khos2ow I think it is fine to remove any unneeded/deprecated features before adding nee tests.

Issue #408, opened 2021-02-26
Anchor support for resources, modules, inputs, outputs, etc


nitrocode:
### What problem are you facing?

I cannot link to a specific variable in the markdown.

### How could terraform-docs help solve your problem?

I'd like each variable to have an anchor associated with it so I can right click, copy link, and share it with someone.

We can do this with the official terraform docs for [resource arguments](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/elasticache_replication_group#transit_encryption_enabled) but not module inputs on hashicorp registry or from terraform-docs.

Here is an [example in markdown from Stackoverflow](https://stackoverflow.com/a/7335259/2965993) of a way to put anchors in markdown

Here is an [example of the outputted anchors](https://gist.github.com/nitrocode/4de07fd583ac864b1d871634c890cc8d#inputs_unquoted) from `table-WithRequired.golden`

I propose anchors prefixed with their section name

* `resources_`
* `modules_`
* `inputs_`
* `outputs_`
* `providers_`

I'm mostly interested in the markdown output since that affects all the git repos. Would we want to do this for other output types as well ? or could that wait for a follow up PR ?

khos2ow:
Thanks @nitrocode, this is a really good suggestion! The only drawback I can think of is that we're going to mix raw html and markdown and any editor and markdown linter would not be happy about this and will show tons of warning messages.

That being said, I'd like to hear more from folks to see if anyone objects and we might want to potentially add an off-switch for this.

nitrocode:
Oh I didn't think that could be an issue. I'll look into adding that a switch this weekend.

nitrocode:
As for `markdownlint` messages

```bash
$ wget https://gist.githubusercontent.com/nitrocode/4de07fd583ac864b1d871634c890cc8d/raw/table-WithRequired.md
$ markdownlint table-WithRequired.md 2>&1 | cut -d' ' -f2 | sort | uniq -c
      2 MD004/ul-style
     36 MD013/line-length
     45 MD033/no-inline-html
      1 MD034/no-bare-urls
      1 MD041/first-line-heading/first-line-h1
      1 MD047/single-trailing-newline
```

* `MD033/no-inline-html` is flagged due to this change
* `MD013/line-length` has probably has increased. Not sure how to lower the length of each line as that would break the table.

By adding the following to the top of the outputted markdown, we can disable those rules from applying but maybe that's not something that should be done.

```html
[...]

PR #409, opened 2021-02-26
Anchors

nitrocode:

Fixes https://github.com/terraform-docs/terraform-docs/issues/408

I have:

Tested this locally and seems to work well with a single test case. Not all test cases pass. Waiting for more discussion.

Is there a way to regenerate the testdata or should I be updating that manually ?

### DIFF
--- a/cmd/asciidoc/asciidoc.go
+++ b/cmd/asciidoc/asciidoc.go
@@ -34,6 +34,7 @@ func NewCommand(config *cli.Config) *cobra.Command {
 	cmd.PersistentFlags().BoolVar(&config.Settings.Required, "required", true, "show Required column or section")
 	cmd.PersistentFlags().BoolVar(&config.Settings.Sensitive, "sensitive", true, "show Sensitive column or section")
 	cmd.PersistentFlags().IntVar(&config.Settings.Indent, "indent", 2, "indention level of AsciiDoc sections [1, 2, 3, 4, 5]")
+	cmd.PersistentFlags().BoolVar(&config.Settings.Anchor, "anchor", true, "create anchor links")
 
 	// subcommands
 	cmd.AddCommand(document.NewCommand(config))
--- a/cmd/markdown/markdown.go
+++ b/cmd/markdown/markdown.go
@@ -35,6 +35,7 @@ func NewCommand(config *cli.Config) *cobra.Command {
 	cmd.PersistentFlags().BoolVar(&config.Settings.Sensitive, "sensitive", true, "show Sensitive column or section")
 	cmd.PersistentFlags().BoolVar(&config.Settings.Escape, "escape", true, "escape special characters")
 	cmd.PersistentFlags().IntVar(&config.Settings.Indent, "indent", 2, "indention level of Markdown sections [1, 2, 3, 4, 5]")
+	cmd.PersistentFlags().BoolVar(&config.Settings.Anchor, "anchor", true, "create anchor links")
 
 	// subcommands
 	cmd.AddCommand(document.NewCommand(config))
--- a/internal/template/anchor.go
+++ b/internal/template/anchor.go
@@ -1,0 +1,42 @@
+/*
+Copyright 2021 The terraform-docs Authors.
+
+Licensed under the MIT license (the "License"); you may not
+use this file except in compliance with the License.
+
+You may obtain a copy of the License at the LICENSE file in
+the root directory of this source tree.
+*/
+
+package template
+
+import (
+	"fmt"
+
+	"github.com/terraform-docs/terraform-docs/internal/print"
+)
+
+// createAnchor
+func createAnchorMarkdown(t string, s string, settings *print.Settings) string {
+	sanitizedName := sanitizeName(s, settings)
+	if settings.ShowAnchor {
+		anchorName := fmt.Sprintf("%s_%s", t, s)
+		sanitizedAnchorName := sanitizeName(anchorName, settings)
+		// the <a> link is purposely not sanitized as this breaks markdown formatting
+		return fmt.Sprintf("<a name=\"%s\"></a> [%s](#%s)", anchorName, sanitizedName, sanitizedAnchorName)
+	}
+
+	return sanitizedName
+}
+
+// createAnchorAsciidoc
+func createAnchorAsciidoc(t string, s string, settings *print.Settings) string {
+	sanitizedName := sanitizeName(s, settings)
+	if settings.ShowAnchor {
+		anchorName := fmt.Sprintf("%s_%s", t, s)
+		sanitizedAnchorName := sanitizeName(anchorName, settings)
+		return fmt.Sprintf("[[%s]] <<%s,%s>>", sanitizedAnchorName, sanitizedAnchorName, sanitizedName)
+	}
+
+	return sanitizedName
+}
--- a/internal/template/anchor_test.go
+++ b/internal/template/anchor_test.go
@@ -1,0 +1,121 @@
+/*
+Copyright 2021 The terraform-docs Authors.
+
+Licensed under the MIT license (the "License"); you may not
+use this file except in compliance with the License.
+
+You may obtain a copy of the License at the LICENSE file in
+the root directory of this source tree.
+*/
+
+package template
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/terraform-docs/terraform-docs/internal/print"
+)
+
+func TestAnchorMarkdown(t *testing.T) {
+	tests := []struct {
+		typeSection string
+		name        string
+		anchor      bool
+		escape      bool
+		expected    string
+	}{
+		{
+			typeSection: "module",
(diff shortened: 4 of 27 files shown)


internal/format/markdown_table.go
```diff
@@ -55,7 +55,7 @@ const (
 			| Name | Version |
 			|------|---------|
 			{{- range .Module.Requirements }}
-				| {{ name .Name }} | {{ tostring .Version | default "n/a" }} |
+				| <a name="requirements_{{ .Name }}"></a> [{{ name .Name }}](#requirements_{{ .Name }}) | {{ tostring .Version | default "n/a" }} |
```
khos2ow, line 58: This is going to be a very (very) repetitive change if you want to do it this way. I'd suggest create new template function (like `anchorName` or something) and move the `<a name=....></a>...` into that. So effectively the change will become:

```diff
-				| {{ name .Name }} | {{ tostring .Version | default "n/a" }} |
+				| {{ anchorName .Name }} | {{ tostring .Version | default "n/a" }} |
```

khos2ow (commented):
I really like the idea @nitrocode and just wanted to give you a very early feedback.

internal/format/markdown_table.go
nitrocode: That's a good point. I'll add a new function for this.

nitrocode: Finished. Thanks for the comment!

khos2ow:
Please hold off any major work on testing until #410 is done.

nitrocode:
Ah I just added the `--anchors` switch and pushed it. I will hold off on making additional changes.

internal/template/template.go
```diff
@@ -58,6 +58,11 @@ func New(settings *print.Settings, items ...*Item) *Template {
 		"sanitizeAsciidocTbl": func(s string) string {
 			return sanitizeItemForAsciidocTable(s, settings)
 		},
+		"anchorName": func(s string, t string) string {
+			copy := *settings
+			s = createAnchor(s, t, &copy)
+			return s
```
khos2ow, line 64: This can be simplified as:

```go
			return createAnchor(s, t, settings)
```

internal/template/anchor.go
```diff
@@ -0,0 +1,31 @@
...
+	var anchor string
+
+	if settings.ShowAnchors {
+		anchorName := fmt.Sprintf("%s_%s", t, s)
+		anchor = fmt.Sprintf("<a name=\"%s\"></a> [%s](#%s)", anchorName, s, anchorName)
+	} else {
+		anchor = s
+	}
+
+	return anchor
```
khos2ow, line 30: This can be simplified as:

```go
	if settings.ShowAnchors {
		anchorName := fmt.Sprintf("%s_%s", t, s)
		return fmt.Sprintf("<a name=\"%s\"></a> [%s](#%s)", anchorName, s, anchorName)
	}

	return s
```

internal/format/testdata/markdown/table-WithRequired.golden
```diff
@@ -40,26 +40,26 @@ followed by another line of text.
 
 | Name | Version |
```
khos2ow, line 41: Instead of modifying `WithRequired` golden file, you can create a new golden file (and its corresponding item in test map for `WithAnchor`. And Please do that for:

- markdown document
- markdown table
- asciidoc document
- asciidoc table

cmd/markdown/markdown.go
```diff
@@ -35,6 +35,7 @@ func NewCommand(config *cli.Config) *cobra.Command {
 	cmd.PersistentFlags().BoolVar(&config.Settings.Sensitive, "sensitive", true, "show Sensitive column or section")
 	cmd.PersistentFlags().BoolVar(&config.Settings.Escape, "escape", true, "escape special characters")
 	cmd.PersistentFlags().IntVar(&config.Settings.Indent, "indent", 2, "indention level of Markdown sections [1, 2, 3, 4, 5]")
+	cmd.PersistentFlags().BoolVar(&config.Settings.Anchors, "anchors", true, "create anchors")
```
khos2ow, line 38: Please also add the same flag to `asciidoc` command too.

khos2ow (commented):
Thank you @nitrocode for starting working on this. This looks ok so far but it needs a bit of changing.

nitrocode:
I'm having issues running test on the `master` branch.

```
$ make test
package github.com/nitrocode/terraform-docs/scripts/docs
	scripts/docs/generate.go:28:2: use of internal package github.com/terraform-docs/terraform-docs/internal/terraform not allowed
```

khos2ow:
I'm wondering why it says `package github.com/nitrocode/terraform-docs/scripts/docs`? Technically you cannot access internal packages from `github.com/terraform-docs/terraform-docs` from `github.com/nitrocode/terraform-docs`.

Are you in GOPATH? If yes, `$GOPATH/src/github.com/terraform-docs/terraform-docs` should be your local cloned folder.

nitrocode:
Thanks. I upgraded golang outside of homebrew and it's working now.

nitrocode:
I was able to fix the commit history by basically deleting my branch and updating the files again... squashing commits before and after an in-between merge does not seem possible.

The asciidoc template doesn't look like it will work currently as anchors are different in asciidoc vs markdown.

Docs
* https://docs.asciidoctor.org/asciidoc/latest/syntax-quick-reference/#links
* https://powerman.name/doc/asciidoc#anchor-1

```asciidoc
[[anchor-1]]
Paragraph or block 1.

anchor:anchor-2[]
Paragraph or block 2.

<<anchor-1>>,
<<anchor-1,First anchor>>,
xref:anchor-2[],
xref:anchor-2[Second anchor].
```

I'll use the above model to update the asciidoc anchors.

nitrocode:
@khos2ow could you review this again please ?

cmd/asciidoc/asciidoc.go
```diff
@@ -34,6 +34,7 @@ func NewCommand(config *cli.Config) *cobra.Command {
 	cmd.PersistentFlags().BoolVar(&config.Settings.Required, "required", true, "show Required column or section")
 	cmd.PersistentFlags().BoolVar(&config.Settings.Sensitive, "sensitive", true, "show Sensitive column or section")
 	cmd.PersistentFlags().IntVar(&config.Settings.Indent, "indent", 2, "indention level of AsciiDoc sections [1, 2, 3, 4, 5]")
+	cmd.PersistentFlags().BoolVar(&config.Settings.Anchors, "anchors", true, "create anchors")
```
khos2ow, line 37: I think singular use of `anchor` is better, throughout the whole PR.

internal/template/anchor_test.go
```diff
@@ -0,0 +1,93 @@
...
+		typeSection string
+		name        string
+		anchors     bool
+		escape      bool
+		expected    string
+	}{
+		{
+			typeSection: "module",
+			name:        "banana_anchors_escape",
+			anchors:     true,
```
khos2ow, line 32: Please add same set of tests for `anchors: false` as well.

khos2ow (commented):
Couple of minor comments.

khos2ow:
- codecov coverage dropped by 0.29% which can be ignored, but you also can add cases into `templates_test.go#TestBuiltinFunc` to cover newly added functions too.
- that's weird. what does `make docs` do, same error? and why do you have `GOOS=linux`? I thought you were using macos.

nitrocode:
Thanks for reviewing. I'll jump on it today.

Yes, I'm on OSX and the issue was related to the `go1.16` issue again...

Not to add too much scope creep to this PR but it would be nice to use an env var in the `Makefile` like `GOBINARY=go` which can be used in every target then if I need to overwrite it, I can simply do either

```
export GOBINARY=go1.16
make docs
```

or

```
GOBINARY=go1.16 make docs
```

khos2ow:
That's a really good point, I'll add it in another PR! :+1:

khos2ow (commented):
One last thing to do, please add `anchor` in:

- `internal/cli/reader.go#L96`
- `docs/reference/config-file.md`

Also please add a bit more information to commit message, as it will be shown in release note.

khos2ow (approved):
Thank you @nitrocode for all the efforts on this PR! :+1:

nitrocode:
I looked at all the tickets for the `v0.12` project and it seems like this won't be released for some time. Would you be willing to cut a `terraform-docs v0.12-beta1` release like they do with terraform ?

khos2ow:
Yeah sure I'll cut a release soon, but also in the meantime you can use `edge` docker image, which is always HEAD of master. (although that might force a change in your workflow, but at least it's an option to immediately unblock you)

nitrocode:
@khos2ow I'm playing with the new release now and it works great!

One thing. On the release notes, you put @ 9trocode instead of @nitrocode 😅

khos2ow:
Interesting! The list of contributors are being extracted automatically, I should check that script to see why that happened! 😅 

Updated release note, thanks for pointing this out and your contribution!

PR #436, opened 2021-03-17
#435: Add pre commit hook

### LINKED ISSUE #435: Add support for pre-commit

As a user
I'd like to be able to run terraform-docs via a [`pre-commit`](https://pre-commit.com/) hook
So that I can easily add this to my existing hooks, use it as a check in a CI pipeline with [this GitHub Action](https://github.com/pre-commit/action), and so forth.

By [adding hook support](https://pre-commit.com/#creating-new-hooks). 😄

tpansino:

Fixes #435

I did find the [existing instructions](https://terraform-docs.io/user-guide/how-to/#git-hook) on how to use a Git hook, and the references to [antonbabenko/pre-commit-terraform](https://github.com/antonbabenko/pre-commit-terraform), but I wanted to provide a pre-commit hook here, on this repo, so that users don't require a second repo just to use this tool via pre-commit.

I have:

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,31 @@
+---
+# Build and run the Go binary from the repo
+# REQUIRES: Go v1.16+ installed
+- id: terraform-docs-go
+  name: terraform-docs
+  description: Generate documentation for Terraform modules (via Go binary)
+  language: golang
+  entry: terraform-docs
+  pass_filenames: false
+  types: [terraform]
+
+# Build and run `terraform-docs` assuming it was installed manually
+# or via package manager
+# REQUIRES: terraform-docs to be installed and on the $PATH
+- id: terraform-docs-system
+  name: terraform-docs
+  description: Generate documentation for Terraform modules (via locally-installed CLI)
+  language: system
+  entry: terraform-docs
+  pass_filenames: false
+  types: [terraform]
+
+# Builds and runs the Docker image from the repo
+# REQUIRES: Docker installed
+- id: terraform-docs-docker
+  name: terraform-docs
+  description: Generate documentation for Terraform modules (via Docker build)
+  language: docker
+  entry: terraform-docs
+  pass_filenames: false
+  types: [terraform]
--- a/docs/user-guide/how-to.md
+++ b/docs/user-guide/how-to.md
@@ -123,6 +123,54 @@ jobs:
 Read more about [terraform-docs GitHub Action] and its configuration and
 examples.
 
+## Pre-commit Hook
+
+With [`pre-commit`], you can ensure your Terraform module documentation is kept
+up-to-date each time you make a commit.
+
+First, simply create or update a `.pre-commit-config.yaml`
+in the root of your Git repo with at least the following content:
+
+```yaml
+repos:
+  - repo: https://github.com/terraform-docs/terraform-docs
+    rev: <VERSION TAG OR SHA TO USE> # For example: "v0.11.2"
+    hooks:
+      - id: terraform-docs-go
+        args: [<ARGS TO PASS INCLUDING PATH>]  # For example: ["--output-file", "README.md", "./mymodule/path"]
+```
+
+(You can also include more than one entry under `hooks:` to update multiple docs.
+Just be sure to adjust the `args:` to pass the path you want terraform-docs to scan.)
+
+Second, install [`pre-commit`] and run `pre-commit` to activate the hooks.
+
+Then, make a Terraform change, `git add` and `git commit`!
+Pre-commit will regenerate your Terraform docs, after which you can
+rerun `git add` and `git commit` to commit the code and doc changes together.
+
+You can also regenerate the docs manually by running `pre-commit -a terraform-docs`.
+
+### Pre-commit via Docker
+
+The pre-commit hook can also be run via Docker, for those who don't have Go installed.
+Just use `id: terraform-docs-docker` in the previous example.
+
+This will build the Docker image from the repo, which can be quite slow.
+To download the pre-built image instead, change your `.pre-commit-config.yaml` to:
+
+```yaml
+repos:
+  - repo: local
+    hooks:
+      - id: terraform-docs
+        name: terraform-docs
+        language: docker_image
+        entry: quay.io/terraform-docs/terraform-docs:latest  # Or, change latest to pin to a specific version
+        args: [<ARGS TO PASS INCLUDING PATH>]  # For example: ["--output-file", "README.md", "./mymodule/path"]
+        pass_filenames: false
+```
+
 ## Git Hook
 
 A simple git hook (`.git/hooks/pre-commit`) added to your local terraform
@@ -146,5 +194,6 @@ Please refer to it for complete examples and guides.
 [sections]: {{< ref "configuration/#sections" >}}
 [output]: {{< ref "configuration/#output" >}}
 [terraform-docs GitHub Action]: https://github.com/terraform-docs/gh-actions
+[`pre-commit`]: https://pre-commit.com/
 [git hooks]: https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks
 [pre-commit-terraform]: https://github.com/antonbabenko/pre-commit-terraform


docs/user-guide/how-to.md
```diff
@@ -123,6 +123,53 @@ jobs:
...
+```yaml
+repos:
+  - repo: git://github.com/terraform-docs/terraform-docs
+    rev: <VERSION TAG OR SHA TO USE> # For example: "v0.11.2"
+    hooks:
+      - id: terraform-docs
+        args: [<ARGS TO PASS INCLUDING PATH>]  # For example: ["--output-file", "README.md", "./mymodule/path"]
+```
+
+(You can also include more than one entry under `hooks:` to update multiple docs. Just be sure
```
khos2ow, line 143: Nit: try to keep the line length roughly at 80 characters, For example `Just be sure` should be moved to next line and the next line gets adjusted accordingly, if need be.

docs/user-guide/how-to.md
```diff
@@ -123,6 +123,53 @@ jobs:
...
+to adjust the `args:` to pass the path you want terraform-docs to scan.)
+
+Second, install [`pre-commit`] and run `pre-commit` to activate the hooks.
+
+Then, make a Terraform change, `git add` and `git commit`! Pre-commit will regenerate your Terraform docs, after which you can
```
khos2ow, line 148: Nit: line length comment as above.

docs/user-guide/how-to.md
```diff
@@ -148,3 +195,5 @@ Please refer to it for complete examples and guides.
 [terraform-docs GitHub Action]: https://github.com/terraform-docs/gh-actions
 [git hooks]: https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks
 [pre-commit-terraform]: https://github.com/antonbabenko/pre-commit-terraform
+[pre-commit]: https://pre-commit.com/
```
khos2ow, line 198:
```suggestion
[`pre-commit`]: https://pre-commit.com/
```

Otherwise this doesn't appear as hyperlink.

docs/user-guide/how-to.md
```diff
@@ -148,3 +195,5 @@ Please refer to it for complete examples and guides.
...
+[release tags]: https://github.com/terraform-docs/terraform-docs/releases
```
khos2ow, line 199: Nit: I'd prefer to keep these footer links in lexicographic order, for easier to find in the future modifications.

docs/user-guide/how-to.md
```diff
@@ -123,6 +123,53 @@ jobs:
 Read more about [terraform-docs GitHub Action] and its configuration and
 examples.
 
+## Pre-commit Hook
+
+With [`pre-commit`], you can ensure your Terraform module documentation is kept
+up-to-date each time you make a commit.
+
+First, simply create or update a `.pre-commit-hooks.yaml`
+in the root of your Git repo with at least the following content:
```
khos2ow, line 132: Nit: `s/Git/git/`

.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,14 @@
+- id: terraform-docs
+  name: terraform-docs
+  description: Generate documentation for Terraform modules
+  language: golang
+  entry: terraform-docs
+  pass_filenames: false
+  types: [terraform]
```
khos2ow, line 7: I'm not too familiar with building pre-commit hooks, does this mean it's going to build the binary on the fly for each execution? How does it know how to build it?
Is there an option to use whatever is installed in `PATH` instead?

khos2ow (commented):
Thank you @tpansino for the work. I have couple of nit comments and a question below.

I'm gonna try this out soon and most probably will come back with more questions!

.pre-commit-hooks.yaml
khos2ow: I spoke too soon! [This](https://pre-commit.com/#golang) is how it works, right?

Also it seems there is a [system](https://pre-commit.com/#system) language. Can we define one additional hook for it? Something like this:

```yaml
- id: terraform-docs
  language: system
  entry: terraform-docs

- id: terraform-docs-go
  language: golang
  entry: terraform-docs

- id: terraform-docs-docker
  language: docker
  entry: terraform-docs
```

docs/user-guide/how-to.md
tpansino: I'm not sure I understand. It matches all of the other links in the block, and those seem to work. Is there something different about mine?

docs/user-guide/how-to.md
tpansino: I considered sorting them, but the existing links don't look sorted to me, so I just added to the bottom.

Are you asking me to sort the entire block? I'm happy to, but to save yourself future reviewing effort I recommend we at least add a comment to say these should be sorted, or better find a linter that will auto-sort and add to the CI pipeline.

docs/user-guide/how-to.md
khos2ow: Yes, whatever goes inside `[ ]` has to match with footer's link. You can check this in action [here](https://deploy-preview-436--terraform-docs.netlify.app/user-guide/how-to/#pre-commit-hook) where instead of hyperlink it gets generated as literal "[`pre-commit`]".

Other option would be to name the link to prevent using backticks, something like:

```markdown
With [`pre-commit`][any-random-id], you can ensure your ...
...

[any-random-id]: https://pre-commit.com/
```

docs/user-guide/how-to.md
tpansino: My understanding is that Git (the project, not the command) should be capitalized here, as it is in the [Git documentation](https://git-scm.com/book/en/v2/Git-Basics-Getting-a-Git-Repository).

I noticed it's not capitalized it most of the documentation though, so I'm happy to change it if consistency is preferred.

.pre-commit-hooks.yaml
tpansino: Yes, that is how it works. Admittedly I cannot get it to build on my box, I get:

```
build github.com/terraform-docs/terraform-docs: cannot load embed: malformed module path "embed": missing dot in first path element
```

Is this something you've seen before? I'm not a Go developer so I just used Docker instead 😅

The `system` hook is generally for unsupported languages or languages that don't have virtual env support. What would be the reason to add that for a Go-based project instead of using the native support?

docs/user-guide/how-to.md
tpansino: Thanks, will fix!

docs/user-guide/how-to.md
khos2ow: Oh I thought they are already sorted, maybe I'm wrong, let me double check.

Anyways I prefer them to be in order, even if they are wrong at the moment. And to be honest I cannot think of a way to automate this and eventually it's not a very big deal if they are not sorted.

docs/user-guide/how-to.md
khos2ow: That's a very good point! Ignore my comment and keep it as `Git`!

tbugfinder:
Please keep in mind that the pre-commt will download this repo although only the pre-commit definition is needed. So an "external" repo makes indeed sense. It could be within the terraform-docs world though.

docs/user-guide/how-to.md
```diff
@@ -123,6 +123,54 @@ jobs:
...
+
+```yaml
+repos:
+  - repo: git://github.com/terraform-docs/terraform-docs
```
tbugfinder, line 136: I'd prefer https protocol.

.pre-commit-hooks.yaml
khos2ow: First of all that error means you don't have go1.16 installed on your machine. Which actually proves my point a bit!

I'm still not sure about best practices of building a git hook, but I imagine most of terraform-docs users don't necessarily have go installed or not the version we depend on (yourself included), but they most probably have terraform-docs installed (with pre-built binaries, brew, or some other way.) So giving folks an option to choose between binary on `PATH` or Docker image or alternatively building from go source.

.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,14 @@
+- id: terraform-docs
```
tbugfinder, line 1: maybe also run yamllint on this file utilizing a pre-commit hook? :-)

docs/user-guide/how-to.md
```diff
@@ -123,6 +123,54 @@ jobs:
...
+## Pre-commit Hook
+
+With [`pre-commit`], you can ensure your Terraform module documentation is kept
+up-to-date each time you make a commit.
+
+First, simply create or update a `.pre-commit-hooks.yaml`
+in the root of your Git repo with at least the following content:
+
+```yaml
+repos:
```
tbugfinder, line 135: yamllint likes `---`.

tbugfinder:
I'm not the maintainer, just sharing my thoughts.

docs/user-guide/how-to.md
khos2ow: something like `https://github.com/terraform-docs/pre-commit-hook` with effectively two files `README` and `.pre-commit-hooks.yml`?

tpansino: Good idea!

.pre-commit-hooks.yaml
tpansino: Indeed that would be great! But I don't think that's really in scope for this PR, it should be something the regular maintainers decide they want to add to the workflow.

docs/user-guide/how-to.md
khos2ow: Cool, so I'm gonna create the repository now. What would you prefer:

- `terraform-docs/pre-commit`
- `terraform-docs/pre-commit-hook`
- anything else?

@tpansino would you be willing to move the `.pre-commit-hooks.yml` file over there and remove it from this PR?

.pre-commit-hooks.yaml
tpansino: > First of all that error means you don't have go1.16 installed on your machine.

Weird, I thought I tried with 1.16, but I just reinstalled to confirm and it works now 🤷

> I imagine most of terraform-docs users don't necessarily have go installed or not the version we depend on

Right, though there are a few ways to handle this:

1. `language_version` parameter on the hook to enforce a minimum version
2. Use Docker
3. `system` hook

If I add `language_version` to the hook file, I get:
```
An unexpected error has occurred: AssertionError: For now, pre-commit requires system-installed golang
```
So I guess that's out 😄 

I suppose adding the `system` is probably the next best option, until Golang 1.16+ is more widely used. The downside is users have one more thing to have to install locally, which goes against the philosophy of pre-commit, but I prefer giving people that choice. 👍

khos2ow: That's a valid point, and since we're considering a standalone repository for the hook itself, why not a third option!

- have `golang` and `docker` hooks reside in `terraform-docs/terraform-docs` (this repo has the code to build anyways)
- have `system` hook reside in `terraform-docs/pre-commit-hook` (specially that this repo won't be able to build at all)

But that might complicates things a bit more than it should be, but we're giving clear choices for folks out there however they prefer to utilize this hook.

docs/user-guide/how-to.md
tpansino: I don't understand - why have a separate repo? That just creates more maintainer overhead for you IMO.

I think @tbugfinder was just asking to switch the proto in the example to https, which doesn't require users to have an SSH set up with GitHub to download the hook. I thought that was pretty reasonable, made the change.

docs/user-guide/how-to.md
tpansino: Sorted now ✔️

docs/user-guide/how-to.md
tbugfinder: Hm, indeed default to https.

On the other hand I do see an option for terraform-docs/pre-commit-hooks (3rd option) as I don't need tool source code just for a hook having few lines.

.pre-commit-hooks.yaml
tpansino: I wouldn't spread it across two repos.

As a pre-commit user, I generally know to look for a `.pre-commit-hooks.yaml` in the repo of a linter I'm considering adopting. The presence of such a file is an immediate adoption boost for me, because I only need to add (for example):

```
  - repo: git://github.com/pre-commit/pre-commit-hooks
    rev: v3.2.0
    hooks:
      - id: end-of-file-fixer
```

To my existing hooks, and boom 💥, I'm already using the tool. I don't install anything extra (except maybe Golang if I don't have it, and even then that's why there's a Docker option), don't have to look across multiple repos, it just works™️.

Most repos I've seen only have one hook, they don't offer a Docker or `system` option. I think it's great that this tool has those options, but it decreases discoverability IMO to spread it out.

tpansino:
> So an "external" repo makes indeed sense. It could be within the terraform-docs world though.

@tbugfinder most repos I've seen include the `.pre-commit-hooks.yaml` as part of their main repo, because it eases adoption and repo maintenance.

Do you have some pre-commit examples where the maintainers keep them separated and it provided benefit?

docs/user-guide/how-to.md
khos2ow: I apologize that I confused you, I should have made my intentions way more clearer! What I meant by sorted, was sort footer links by where they appear in the document. For example `[sections]` which is first defined in line 13 should be placed in footer before `[output]` which is first defined in line 75 and so on.

tbugfinder:
> > So an "external" repo makes indeed sense. It could be within the terraform-docs world though.
> 
> @tbugfinder most repos I've seen include the `.pre-commit-hooks.yaml` as part of their main repo, because it eases adoption and repo maintenance.
> 
> Do you have some pre-commit examples where the maintainers keep them separated and it provided benefit?

I checked a few projects I'm currently using - (I cannot review all GitHub projectsby end of the week.). Apart from those projects which do not have pre-commit-hooks at all, there are more repos than expected which indeed provide the pre-commit configuration in the same repo as the source code.

docs/user-guide/how-to.md
khos2ow: I think after this discussion, we can have the approaches mentioned at https://github.com/terraform-docs/terraform-docs/pull/436#discussion_r596352055, and eventually (outside of scope of this PR) update docs to point them out. So folks can effectively choose between `golang`, `docker`, or `system` hook based on their requirements or liking.

docs/user-guide/how-to.md
tpansino: > I apologize that I confused you, I should have made my intentions way more clearer!

You did say "in lexicographic order", which I took her to mean "alphabetically". I take it what you meant to say was "in order of first appearance".

That strikes me as an odd system though, given that, if I want to find the link for pre-commit, and it happens to have appeared earlier in the document, I might not know, and have to dig for it in the footer links. Alphabetical fixes that case, and is what I'd expect a linter/formatter tool to do, but I can change it back to how it was.

khos2ow: That's a good point, then leave as is (alphabetically sorted) and we can see if that proves problematic in the future and we'll modify later. 👍

docs/user-guide/how-to.md
tpansino: So I currently have pushed up all three hook options. They should all work as-is in this repo. 

Can I mark this resolved then? It sounds like creating a separate repo for the `system` hook is a future improvement?

.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,29 @@
...
+# Build and run the Go binary from the repo
+# REQUIRES: Go v1.16+ installed
+- id: terraform-docs-go
+  name: terraform-docs
+  description: Generate documentation for Terraform modules (via Go binary)
+  language: golang
+  entry: terraform-docs
+  pass_filenames: false
+  types: [terraform]
+# Build and run `terraform-docs` assuming it was installed manually
```
khos2ow, line 11: Nit: please put an empty line between these blocks for better readability.

khos2ow (commented):
One small comment below, and after our discussions I believe we can proceed with this PR and think about standalone repository only for `system` hook later on.

Also please squash your commits.

docs/user-guide/how-to.md
tpansino: Further comment 🙂 - adding links like this is quite a pain for a contributor, as I have to search through the entire page to find which if any link came before it, and then jump back to the bottom and insert my new link. And as I move paragraphs around and refactor my doc I have to adjust the ordering of the links.

It's not great UX for either the reader, contributor, or maintainer IMO, but I've added it correctly for consistency.

docs/user-guide/how-to.md
khos2ow: I think it's best in not mark these as resolved, for folks looking at this PR in the future to see the discussion (although logically they are resolved)

docs/user-guide/how-to.md
```diff
@@ -123,6 +123,54 @@ jobs:
 Read more about [terraform-docs GitHub Action] and its configuration and
 examples.
 
+## Pre-commit Hook
+
+With [`pre-commit`], you can ensure your Terraform module documentation is kept
+up-to-date each time you make a commit.
+
+First, simply create or update a `.pre-commit-hooks.yaml`
```
khos2ow, line 131: Is this a typo? I think this should be `.pre-commit-config.yaml` for consumers of this git-hook.

tpansino: Good catch! Thank you!

.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,31 @@
+---
+# Build and run the Go binary from the repo
+# REQUIRES: Go v1.16+ installed
+- id: terraform-docs-go
```
khos2ow, line 4: `s/terraform-docs-go/terraform-docs/` to match the example yaml down below.

khos2ow (commented):
I promise this was the last one!! :sweat_smile:

.pre-commit-hooks.yaml
tpansino: I intentionally left this one. I think if you're going to offer multiple hooks that do the same thing via different platforms, then there needs to be a clear distinction that the consumer needs to choose one.

I also think making the `id: terraform-docs` implies this is the reasonable default, which we've established it's not for the Go-binary option because most people don't have v1.16 or higher installed. At best, I'd make the default the Docker option right now, but I think that's inconsistent with most other pre-commit hooks I've seen, hence including the suffix on all three.

khos2ow: Fair point, then can you update example to be using go hook. Because the example as it stands is broken.

tpansino:
> Also please squash your commits.

Why do the contributing guidelines say:

```
Don't force push to address review feedback. Your commits should tell a story.
```

If you want me to squash? That seems contradictory, no?

---

Also @khos2ow, some polite feedback: while I respect the maintainer power to Nit and keep things tidy (and indeed, use it myself), I think you need to be WAY more up-front on your expectations. I read the Contributing guidelines, I wrote documentation following the existing style I found. If something needs to be alphabetized, whitespace added for readability, etc, that needs to be either (A) documented in the guidelines or (B) preferably a linter suite I can run. I maintain formal documentation as part of my job, and if I have a style guideline that isn't captured somewhere, then I don't enforce it on my contributors to follow it.

Expectations need to be clear, anything less wastes time with rework. 🙏

khos2ow:
> If you want me to squash? That seems contradictory, no?

The reason we ask not to force push for addressing review comments is for the maintainer to be able only see the recent changes. For example I can only look at fab6de5. But when everything is reviewed and addressed we would like to have a clean(er) commit history on main branch. It can come to PR author discretion to sqash all the commits or rebase interactively and group them logically together. For example:

1. `Add support for pre-commit hooks`
1. `Update pre-commit documentation`

> Expectations need to be clear, anything less wastes time with rework.

This is a fantastic feedback, I really appreciate you brought it up and we'll definitely try to incorporate them as much as we can in our guidelines! :+1:

khos2ow (approved):
LGTM, Thank you @tpansino for the awesome work!

tpansino:
> The reason we ask not to force push for addressing review comments is for the maintainer to be able only see the recent changes.

Understandable. 

Is there a reason you don't just enable Squash Merge in the repo settings? It does what you're asking without losing the history and context of PR changes, and doesn't require an extra step from contributors.

khos2ow:
> Is there a reason you don't just enable Squash Merge in the repo settings?

The main reason is that, IMO, merging with commits gives a better commit history on the main branch, specially when you want to debug and bisect. And at the same time give contributors to logically group their commits on their PRs however they see fit. For example:

- commit 1: `Add feature X with some information`
- commit 2: `Make sure feature X is really performant`
- commit 3: `Update documentation for feature X`

This, admittedly, might impose some extra work for contributors when reviewing the PR is finished (exactly like here), so you - as a contributor - would have a choice to interactively rebase all the commits and group them together to your liking, or just sqaush everything up.

Anyways, I hope that clarifies a bit of what our intentions were and again I apologize for the confusion our contributing guide might have caused you. :)

PR #450, opened 2021-03-27
print cli errors to stderr

asobrien:

This is a small change that prints any errors that surface when
executing the cli (i.e., `cmd.Execute()`) to stderr rather than
stdout.

When redirecting streams, its useful to keep errors separated from
program output so that in the case of an error the message won't be
redirected and is still surfaced. This issue came up for me when
using this terraform-docs in a script, where I encountered an error:

    $ terraform-docs markdown path/not/to/mod > outfile
    zsh: exit 1     terraform-docs markdown path/not/to/mod > outfile

When I ran the above command in my script an error was generated
because the path was incorrect, however the error message wasn't
printed to my console and the outfile was filled with the error
message. The change here modifies the cli's output to instead show:

    $ terraform-docs markdown path/not/to/mod > outfile
    Error: Failed to read module directory: Module directory path/not/to/mod does not exist or cannot be read.
    zsh: exit 1     terraform-docs markdown path/not/to/mod > outfile

and the outfile is empty.

I have:

This change has been tested by compiling and running an executable with the changes. The change to the console output, as shown above, was generated by running the binary which contained the proposed changes.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -12,6 +12,7 @@ package cmd
 
 import (
 	"fmt"
+	"os"
 
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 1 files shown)


khos2ow (approved):
This is neat! Thanks @asobrien, lgtm!

PR #458, opened 2021-04-03
Add output-check flag option

### LINKED ISSUE #456: Return name of file when using file output

Using GitLab CI to check if terraform docs are up to date. There is no way to know if running `terraform-docs` updated a file when run. 

Using `terraform fmt` as an example, we can easily check if a file changed and fail the pipeline if true. 

When a file changed, returning the filename would allow us to know if a file changed and fail if true. Meaning the docs were not updated when the module changed.

rickhlx:

- Add a `--output-check` flag that only works when outputting to a file.
   - Exit code 0 when no changes
   - Exit code 1 when changes found

Also: 
- Return filename when output file updated (or differences found)

Fixes #456 

I have:

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,10 @@
 
 ![terraform-docs-teaser](./images/terraform-docs-teaser.png)
 
+Sponsored by [Scalr - Terraform Automation & Collaboration Software](https://scalr.com/?utm_source=terraform-docs)
+
+<a href="https://www.scalr.com/?utm_source=terraform-docs" target="_blank"><img src="https://bit.ly/2T7Qm3U" alt="Scalr - Terraform Automation & Collaboration Software" width="175" height="40" /></a>
+
 ## What is terraform-docs
 
 A utility to generate documentation from Terraform modules in various output formats.


cmd/root.go
```diff
@@ -66,6 +66,7 @@ func NewCommand() *cobra.Command {
 	cmd.PersistentFlags().StringVar(&config.Output.File, "output-file", "", "File in module directory to insert output into (default \"\")")
 	cmd.PersistentFlags().StringVar(&config.Output.Mode, "output-mode", "inject", "Output to file method ["+cli.OutputModes+"]")
 	cmd.PersistentFlags().StringVar(&config.Output.Template, "output-template", cli.OutputTemplate, "Output template")
+	cmd.PersistentFlags().BoolVar(&config.Output.Check, "output-check", false, "check if output is generated doc (default false)")
```
khos2ow, line 69: I think this is bit vague. Maybe something like:

```text
Check if content of output file is up to date
```

khos2ow (commented):
Sorry for the delay, can you please check my comments below and also rebase on master?

cmd/root.go
rickhlx: Sounds good to me

rickhlx:
@khos2ow implemented the changes suggested, thanks for reviewing!

internal/cli/config.go
```diff
@@ -187,6 +187,8 @@ const (
 	outputBeginComment = "<!-- BEGIN_TF_DOCS -->"
 	outputContent      = "{{ .Content }}"
 	outputEndComment   = "<!-- END_TF_DOCS -->"
+
+	outputCheck = false
```
khos2ow, line 191: This too should be removed. Make sure to run `make lint` locally, or check the [output](https://github.com/terraform-docs/terraform-docs/pull/458/checks?check_run_id=2504783389) of GitHub Action.

.editorconfig
```diff
@@ -18,6 +18,7 @@ indent_size = 4
 
 [*.md]
 trim_trailing_whitespace = true
+indent_size = 2
```
khos2ow, line 21: To be fair, this is very opinionated. Please revert it back.

rickhlx:
Thanks for the comments @khos2ow, fixed.

internal/cli/writer.go
```diff
@@ -157,5 +159,26 @@ func (fw *fileWriter) write(filename string, p []byte) (int, error) {
 	if fw.writer != nil {
```
khos2ow, line 159: I suggest converting the whole function as follow:

```go
func (fw *fileWriter) write(filename string, p []byte) (int, error) {
	// if run in check mode return exit 1
	if fw.check {
		f, err := os.ReadFile(filename)
		if err != nil {
			return 0, err
		}

		// check for changes and print changed file
		if !bytes.Equal(f, p) {
			return 0, fmt.Errorf("%s is out of date", filename)
		}

		fmt.Printf("%s is up to date.\n", filename)
		return 0, nil
	}

	if fw.writer != nil {
		return fw.writer.Write(p)
	}

	fmt.Printf("%s updated successfully.\n", filename)
	return len(p), os.WriteFile(filename, p, 0644)
}
```

khos2ow: Consecutively this test case can also be added in `writer_test.go` at the end of `TestFileWriter`:

```go
		"ModeReplaceOutOfDate": {
			file:     "mode-replace.md",
			mode:     "replace",
			check:    true,
			template: OutputTemplate,
			begin:    outputBeginComment,
			end:      outputEndComment,
			writer:   nil,

			expected: "",
			wantErr:  true,
			errMsg:   "testdata/writer/mode-replace.md is out of date",
		},
```

Note that you also need to add `check bool` to the struct and `check: false,` to all the other test cases.

khos2ow: @rickhlx did you want to finish off your changes here? or would you rather me adding these new changes instead? Either way is great!

rickhlx: Sorry! Been running around a bit. I'll take care of these changes, no problem. Thanks for the comments!

khos2ow:
@rickhlx I accidentally pushed to the wrong branch and closed this PR! I apologize. I'm gonna open another PR with your changes.

Issue #461, opened 2021-04-08
Add An Option for "tfvars hcl" command to Show Description on the variable


mikiya771:
### What problem are you facing?

When:  Creating Terraform Template,
What:  I want to create tfvars from "variables.tf" with description for my template user.

I want to give my template users an experience that they can write tfvars only reading tfvars (not with variables.tf).

#### Example
input 
```
variables "some_var" {
  default = "test"
  description = "this is a test variable"
}
variables "some_var2" {
  default = "test"
}
```
output
```
/*
this is a test variable
*/
some_var = "test"

# no descriptions
some_var2 = "test"
```
full goal image: https://github.com/mikiya771/terraform-docs/blob/master/internal/format/testdata/tfvars/hcl-PrintDescription.golden

### How could terraform-docs help solve your problem?

I want to get the expected tfvars with an additional option like this.

```
terraform-docs tfvars hcl --description .
```

khos2ow:
That's an interesting suggestion, although I think it can be done with couple of cosmetic changes:

1. use `#` comment instead of `/* */`
2. don't show anything if there's no description

```hcl
# this is a test variable
# with multiple lines
some_var = "test"

some_var2                = "test"
some_var3_with_long_name = "test"

# this is another test variable
some_var4 = "test"
```

mikiya771:
@khos2ow Thank you for your quick response.
I agree with your two advices,

PR #466, opened 2021-04-11
Feature/lssue461/add description flag

mikiya771:

- Add `--description` boolean flag that only works when `tfvars hcl` command.
 

feature #461 

I have:

- Add New Test Pattern  for `trvars hcl` command as `internal/format/testdata/tfvars/hcl-PrintDescription.golden`

### DIFF
--- a/internal/format/testdata/tfvars/hcl-PrintDescription.golden
+++ b/internal/format/testdata/tfvars/hcl-PrintDescription.golden
@@ -1,0 +1,104 @@
+
+# A variable with underscores.
+input_with_underscores = ""
+
+# It's list number two.
+list-2 = ""
+
+# It's map number two.
+map-2 = ""
+
+# It's number number two.
+number-2 = ""
+
+# It's string number two.
+string-2 = ""
+
+string_no_default = ""
+unquoted          = ""
+
+# It's bool number one.
+bool-1 = true
+
+# It's bool number two.
+bool-2 = false
+
+bool-3             = true
+bool_default_false = false
+
+# This is a complicated one. We need a newline.  
+# And an example in a code block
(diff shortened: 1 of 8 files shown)


internal/format/testdata/tfvars/hcl-PrintDescription.golden
```diff
@@ -0,0 +1,104 @@
...
+map-2                  = ""
+
+# It's number number two.
+number-2               = ""
+
+# It's string number two.
+string-2               = ""
+
+string_no_default      = ""
+unquoted               = ""
```
khos2ow, line 18: Description comments are completely fine here, but the whole file is not terraform formatted properly. It should like like something like this:

```hcl
# A variable with underscores.
input_with_underscores = ""

# It's list number two.
list-2 = ""

# It's map number two.
map-2 = ""

# It's number number two.
number-2 = ""

# It's string number two.
string-2 = ""

string_no_default = ""
unquoted          = ""

...
```

khos2ow (commented):
Thank you @mikiya771 for working on this, on a quick look there's one formatting problem stood out, please check my comment below. Essentially when we generate `tfvars hcl` format, the result should be "terraform formatted" properly.

internal/format/testdata/tfvars/hcl-PrintDescription.golden
mikiya771: @khos2ow Thank you for your comment and this formatted example.
Now I fix the code to generate terraform-formatted tfvars.

khos2ow (approved):
Thank you @mikiya771, lgtm. Can you please sqaush the commits?

PR #498, opened 2021-05-06
fix(495): Fix the malformed version by removing the commitHash check.

### LINKED ISSUE #495: Malformed version error in 0.13.0
Hello maintainers!

### Describe the bug

Using latest version 0.13.0 (4354602), and taking a version constraint from the example configuration reference:

`version: ">= 0.13.0, < 1.0.0"`

yields: `Error: Malformed version: 0.13.0 4354602`

I would expect the version constraint be matched if using a correct version (as I am), or an informative error be given if the version I have installed does not meet constraints.

### How can we reproduce it?

To reproduce, add the version constraint to .terraform-docs.yml and try to generate any docs.

### Environment information

terraform-docs version 0.13.0 4354602 linux/amd64 BuildDate: 2021-04-30T20:31:56Z
Ubuntu 20.04
Installed via pre-built binary.

jsrice7391:
Signed-off-by: Justin Rice <jsrice7391@gmail.com>

# Fix the version error.

Right now there are issues with the version within homebrew. The top example is the result of running `--version` in a local build of `terraform-docs` and the second is `--version` on the version currently published (0.13.0).

```bash
$ ./../OpenSource/terraform-docs/terraform-docs --version
terraform-docs version v0.13.0 darwin/amd64 BuildDate: 2021-05-05T23:59:01-04:00
$ terraform-docs --version 
terraform-docs version 0.13.0 4354602 darwin/amd64 BuildDate: 2021-04-30T20:31:56Z
```

This difference is due to `commitHash` not being appended. In the local version, built right from source, it does not have a commit hash baked into the build and therefore does not append.

If we remove the `commitHash` check, we get the same resulting version.

```bash
$  ../../OpenSource/terraform-docs/terraform-docs markdown document --config "$(pwd)/.terraform-docs.yml" modules/s3_redirect
$ terraform-docs markdown document --config "$(pwd)/.terraform-docs.yml" modules/s3_redirect         
Error: Malformed version: 0.13.0 4354602
```

Fixes #495 

I have:

- [X] Read and followed terraform-docs' [contribution process].
- [X] All tests pass when I run `make test`.

### DIFF
--- a/internal/version/version.go
+++ b/internal/version/version.go
@@ -49,6 +49,10 @@ func Short() string {
 
 // Full return the full version of the binary including commit hash and build date
 func Full() string {
+	fullVersion := version
+	if !strings.HasSuffix(version, commitHash) {
+		fullVersion += " " + commitHash
+	}
 	osArch := runtime.GOOS + "/" + runtime.GOARCH
-	return fmt.Sprintf("%s %s BuildDate: %s", version, osArch, buildDate)
+	return fmt.Sprintf("%s %s BuildDate: %s", fullVersion, osArch, buildDate)
 }
--- a/internal/version/version.go
+++ b/internal/version/version.go
@@ -13,7 +13,6 @@ package version
 import (
 	"fmt"
 	"runtime"
-	"strings"
 	"time"
 )
 
@@ -50,9 +49,6 @@ func Short() string {
 
 // Full return the full version of the binary including commit hash and build date
 func Full() string {
-	if !strings.HasSuffix(version, commitHash) {
-		version += " " + commitHash
-	}
 	osArch := runtime.GOOS + "/" + runtime.GOARCH
 	return fmt.Sprintf("%s %s BuildDate: %s", version, osArch, buildDate)
 }


khos2ow (changes requested):
Please check my comments below.

internal/version/version.go
```diff
@@ -50,9 +49,6 @@ func Short() string {
 
 // Full return the full version of the binary including commit hash and build date
 func Full() string {
-	if !strings.HasSuffix(version, commitHash) {
-		version += " " + commitHash
-	}
```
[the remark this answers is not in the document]
jsrice7391: Ahh, that makes sense. I will dive more into go and help with this project. We use a lot of TF, but auto documenting is definitely needed. Awesome work on this. Changes addressed in https://github.com/terraform-docs/terraform-docs/pull/498/commits/70e62fe74ea6dc2886753d01b95859190c6b43ac

internal/version/version.go
```diff
@@ -13,7 +13,6 @@ package version
 import (
 	"fmt"
 	"runtime"
-	"strings"
```
khos2ow, line 16: This needs to be reverted back, it's being used.

khos2ow (commented):
See my comment below. Also it'd be nice if we add tests for `version.go` as well.

internal/version/version.go
khos2ow: @jsrice7391 did you want to finish off your changes here? or would you rather me adding them? Either way is great!

jsrice7391: Unfortunately no. I have a lot of other things going on, thanks for letting me hack at it! Awesome tool.

khos2ow: For sure, no worries, thanks!

PR #546, opened 2021-08-06
Docker run command was missing critical info to generate documentation.

intikhabalam:
Docker command was incomplete and not generating the documentation. I have added the options to generate the tags.

I have:

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,11 +90,24 @@ choco install terraform-docs
 Alternatively you also can run `terraform-docs` as a container:
 
 ```bash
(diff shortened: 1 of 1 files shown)


flatiron32:
This change still does not produce meaningful results because the container does not have access to the files on the host. Updating it to something like `docker run -v $PWD:/tmp/terraform quay.io/terraform-docs/terraform-docs /tmp/terraform` would be better IMHO.

khos2ow:
Thank you @intikhabalam, going to close this in favor of #561.

PR #554, opened 2021-09-07
introduce exec template func

FalcoSuessgott:

Hi, 

I would like to hear your feedback about a feature that may is interesting for other people.

I made a short PoC for the `exec` template func, which allows you to render output in your templates:

```
I want to have echo output in here: {{ exec "echo test" }}
```

would result in:

```
I want to have echo output in here: `test`
```

There are still some things to take care of, for example quoatation and which shell to use, but I wanted to hear your opionion first.

One use case for me is that I use the `{{ include }}` feature to show cliends how to use a terraform module, however I have to manually bump to `?ref` version part of a terraform module source to the latest tag. With the `exec` template func we could to something like:

```hcl
source = "git@uri?ref={{ exec \"git describe --tags --abbrev=0\" }}"
```

```hcl
source = "git@uri?ref=v0.15.0"
```

Let me hear what u think.

I have:


internal/exec/exec.go
```diff
@@ -0,0 +1,28 @@
+package exec
```
khos2ow, line 1: Leading license comments is missing here (and also in `exec_test.go`), you can copy it over from another `.go` file.

internal/exec/exec.go
```diff
@@ -0,0 +1,28 @@
...
+func RunCommand(command string) (string, error) {
+	shell := getShell()
+
+	output, err := exec.Command(shell, "-c", command).Output()
+	if err != nil {
+		return "", err
+	}
+
+	return string(output), nil
+}
```
khos2ow, line 28: This file is not go formatted properly. Please run `make fmt` locally (you might need to run `make tools` first to install required toolings)

khos2ow (commented):
Thanks @FalcoSuessgott, I really like this idea! Although there are some failure on the CI, can you run `make fmt`, `make lint`, and `make test` locally and fix the issues. (you might want to run `make tools` to install required toolings beforehand)

khos2ow:
Please take a look at [contributing guide](https://github.com/terraform-docs/terraform-docs/blob/master/CONTRIBUTING.md) in regards to commits sign-off. DCO is failing commits sign-off are missing.

PR #564, opened 2021-09-16
docs: typo in README

bcdady:
Adjust typo, and suggest possible readability improvement, in README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,7 +165,7 @@ when you make a commit.
 ## Configuration
 
 terraform-docs can be configured with a yaml file. The default name of this file is
-`.terraform-docs.yml` and the path order for locating for it is:
+`.terraform-docs.yml` and the path order for locating it is:
 
 1. root of module directory
 1. `.config/` folder at root of module directory
--- a/README.md
+++ b/README.md
@@ -164,8 +164,8 @@ when you make a commit.
 
 ## Configuration
 
-terraform-docs can be configured with a yaml file. Default name of this file is
-`.terraform-docs.yml` and the order for lookig for it is:
+terraform-docs can be configured with a yaml file. The default name of this file is
+`.terraform-docs.yml` and the path order for locating for it is:
 
 1. root of module directory
 1. `.config/` folder at root of module directory
--- a/docs/user-guide/configuration.md
+++ b/docs/user-guide/configuration.md
@@ -12,7 +12,23 @@ toc: true
 ---
 
 The `terraform-docs` configuration file uses the [yaml format](https://yaml.org/) in order to override any default behaviors.
-. This is a convenient way to share the configuation amongst teammates, CI, or other toolings.
+This is a convenient way to share the configuration amongst teammates, CI, or other toolings.
+
+terraform-docs will locate any available configuration file without needing to explicitly pass the `--config` flag.
+
+The default name of the configuration file is `.terraform-docs.yml`.
+The path order for locating it is:
+
+1. root of module directory
+1. `.config/` folder at root of module directory <sup class="no-top">(since v0.15.0)</sup>
+1. current directory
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -164,8 +164,8 @@ when you make a commit.
 
 ## Configuration
 
-terraform-docs can be configured with a yaml file. Default name of this file is
-`.terraform-docs.yml` and the order for lookig for it is:
+terraform-docs can be configured with a yaml file. The default name of this file is
+`.terraform-docs.yml` and the path order for locating for it is:
```
khos2ow, line 168: I think "for" after "locating" should be dropped now.

README.md
```diff
@@ -164,8 +164,8 @@ when you make a commit.
 
 ## Configuration
 
-terraform-docs can be configured with a yaml file. Default name of this file is
-`.terraform-docs.yml` and the order for lookig for it is:
+terraform-docs can be configured with a yaml file. The default name of this file is
```
khos2ow, line 167: I like the new wording, and I think it might be good to update https://github.com/terraform-docs/terraform-docs/blob/master/docs/user-guide/configuration.md in similar way.

khos2ow (commented):
Nice catch @bcdady, please see my comments below.

Also unfortunately GitHub UI doesn't give an option to sign-off a commit and that has to be done locally. Can you please sign-off with `git commit -s` and push again?

README.md
bcdady: @khos2ow I edited configuration.md per your suggestion and welcome your feedback

README.md
bcdady: I agree. I think this suggestion/thread can now be resolved.

docs/user-guide/configuration.md
```diff
@@ -29,6 +25,21 @@ $ tree
 $ terraform-docs .
 ```
 
+As of `v0.13.0`, The default name of the configuration file is `.terraform-docs.yml`.
```
khos2ow, line 28:
```
As of `v0.13.0`, The default name of the configuration ...
```

The new change here is misleading, it gives the impression that the default name of the config file has been added i `v0.13.0`, but the ordering was added in v0.13.0.

docs/user-guide/configuration.md
```diff
@@ -11,12 +11,8 @@ weight: 120
 toc: true
 ---
 
-The `terraform-docs` configuration is a yaml file. This is a convenient way to
-share the configuation amongst teammates, CI, or other toolings. To do so you
-can use `-c` or `--config` flag which accepts name of the config file.
-
-Default name of this file is `.terraform-docs.yml`, and it will get picked up
-(if existed) without needing to explicitly passing with config flag.
```
khos2ow, line 19: This is an important line that got dropped as part of this change.

khos2ow: I think this might be better

```suggestion
The default name of this file is `.terraform-docs.yml`, and it will get picked up
(if existed) without needing to explicitly passing with config flag.
```

docs/user-guide/configuration.md
```diff
@@ -42,17 +53,6 @@ $ tree
 $ terraform-docs -c .tfdocs-config.yml .
 ```
 
-As of `v0.13.0`, the order for looking for config file is:
```
khos2ow, line 45: I think it might be better to restore this back, because the logical importance of this section is lower than the previous paragraph on how to use alternative name for config file.

bcdady: ok I can work on addressing your latest feedback today

bcdady: Thanks for clarifying the implied emphasis.
How do these latest edits read to you?

khos2ow (approved):
lgtm!

Issue #550, opened 2021-08-18
Extract all comments, not just header and footer


jpdoria:
### What problem are you facing?

I’m facing a challenge at the moment. Say I have this `project.tf` and I want to extract the comments for each resource or module. Is that possible? How do I do that? Here’s how my `project.tf` looks like:

```hcl
/*
 * # Main Title
 * This is my project
*/

/*
 * ## example._example
 * this resource is blahblah
 *
*/

resource "example" "_example"...

/*
 * ## module.hello
 * this module says hello
 *
*/

module "hello" ...
```

The expected output is:

```markdown
# Main Title

## example._example
this resource is blahblah

## module.hello
this module says hello
```

### How could terraform-docs help solve your problem?

It would be great if terraform-docs can extract all comments, not just the header and footer.

khos2ow:
@jpdoria #569 added the support for reading leading comments for `module` and `resource` as well. To follow on the same convention we had the leading comments should be either:

- `// this is a comment, can be multiline`
- `# this is a comment, can be multiline`

So your example code can be slightly modified to:

```hcl
/*
 * # Main Title
 * This is my project
*/

// ## example._example
// this resource is blahblah
resource "example" "_example"...

// ## module.hello
// this module says hello
module "hello" ...
```

PR #565, opened 2021-09-19
optionally process comments from tf files as description

### LINKED ISSUE #552: Do not use comments as variable description
### Describe the bug

I have a variables.tf file with `# some comment` and then variables definitions follow
When I generate docs, the comment is used as variable description. I want to avoid that. Is there a config setting I can use to prevent this ?

### How can we reproduce it?

1. Create `variables.tf` with definition
```
# some comment
variable "var1" {
}
```
2. Run 
```
▶ terraform-docs markdown --output-check --show inputs ./
```
Input section will be rendered.

 Output is as follows:
```

## Inputs

| Name | Description | Type | Default | Required |
|------|-------------|------|---------|:--------:|
| <a name="input_var1"></a> [var1](#input\_var1) | some comment | `any` | n/a | yes |
```
## Inputs

| Name | Description | Type | Default | Required |
|------|-------------|------|---------|:--------:|
| <a name="input_var1"></a> [var1](#input\_var1) | some comment | `any` | n/a | yes |

## Inputs

| Name | Description | Type | Default | Required |
|------|-------------|------|---------|:--------:|
| <a name="input_var1"></a> [var1](#input\_var1) | n/a | `any` | n/a | yes |

### Environment information

- terraform-docs version (use `terraform-docs --version`):
```
▶ terraform-docs version                                          

terraform-docs version v0.15.0 darwin/amd64
```
- OS (e.g. Windows, MacOS): MacOS

poojitha-bikki:

Fixes #552

This PR includes changes to optionally process comments from `.tf` files as description (when description is not provided in terraform inputs, outputs definition)

- Adds flag **`read-comments`** 

I have:

- Added test `TestReadComments()`
- Tested with `make test`
- Tested manually 

> **Manual Test**
> - Create sample tf folder and file
> ```
> mkdir /tmp/test-tf
> cat <<EOF >> /tmp/test-tf/sample.tf
> # some comment
> variable "var1" {
> }
> 
> # some comment
> output "output-var1" {
>   value = null
> }
> EOF
> ```
> 
> - Tested with read-comments as false. Expected - description should be n/a (should not use comment (#some comment) as description)
> ```go
> ▶ go run main.go --read-comments=false --show=inputs,outputs markdown /tmp/test-tf
> ```
> **Result:**
> ## Inputs
> 
> | Name | Description | Type | Default | Required |
> |------|-------------|------|---------|:--------:|
> | <a name="input_var1"></a> [var1](#input\_var1) | n/a | `any` | n/a | yes |
> 
> ## Outputs
> 
> | Name | Description |
> |------|-------------|
> | <a name="output_output-var1"></a> [output-var1](#output\_output-var1) | n/a |
> 
> - Tested with read-comments as true. Expected - 'some comment' is used as description
> ```
> ▶ go run main.go --read-comments=true --show=inputs,outputs markdown /tmp/test-tf  
> ```
> **Result:**
> ## Inputs
> 
> | Name | Description | Type | Default | Required |
> |------|-------------|------|---------|:--------:|
> | <a name="input_var1"></a> [var1](#input\_var1) | some comment | `any` | n/a | yes |
> 
> ## Outputs
> 
> | Name | Description |
> |------|-------------|
> | <a name="output_output-var1"></a> [output-var1](#output\_output-var1) | some comment |
> - Cleanup
> ```
> $ rm -r /tmp/test-tf
> ```

**TBD**

- ~Config setting that maps to the new option added~ Added new config setting `read-comments`
- ~Tests when option is set to `false`. Currently few tests fail when option is set to `false`~ set default to true (for backwards compatibility). Added new test cases to test when set to `false` and `true`

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -81,6 +81,8 @@ func NewCommand() *cobra.Command {
 	cmd.PersistentFlags().BoolVar(&config.OutputValues.Enabled, "output-values", false, "inject output values into outputs (default false)")
 	cmd.PersistentFlags().StringVar(&config.OutputValues.From, "output-values-from", "", "inject output values from file into outputs (default \"\")")
 
+	cmd.PersistentFlags().BoolVar(&config.Settings.ReadComments, "read-comments", true, "use comments as description when description is empty")
+
 	// formatter subcommands
 	cmd.AddCommand(asciidoc.NewCommand(runtime, config))
 	cmd.AddCommand(json.NewCommand(runtime, config))
--- a/internal/cli/config.go
+++ b/internal/cli/config.go
@@ -40,15 +40,16 @@ var flagMappings = map[string]string{
 	"sort-by-required": "required",
 	"sort-by-type":     "type",
 
-	"anchor":      "settings.anchor",
-	"color":       "settings.color",
-	"default":     "settings.default",
-	"description": "settings.description",
-	"escape":      "settings.escape",
-	"indent":      "settings.indent",
-	"required":    "settings.required",
-	"sensitive":   "settings.sensitive",
-	"type":        "settings.type",
+	"anchor":        "settings.anchor",
+	"color":         "settings.color",
+	"default":       "settings.default",
+	"description":   "settings.description",
+	"escape":        "settings.escape",
+	"indent":        "settings.indent",
+	"read-comments": "settings.read-comments",
+	"required":      "settings.required",
+	"sensitive":     "settings.sensitive",
+	"type":          "settings.type",
 }
 
 // Config represents all the available config options that can be accessed and passed through CLI
@@ -376,34 +377,36 @@ func (s *sort) validate() error {
 }
 
 type settings struct {
-	Anchor      bool `mapstructure:"anchor"`
-	Color       bool `mapstructure:"color"`
-	Default     bool `mapstructure:"default"`
-	Description bool `mapstructure:"description"`
-	Escape      bool `mapstructure:"escape"`
-	HideEmpty   bool `mapstructure:"hide-empty"`
-	HTML        bool `mapstructure:"html"`
-	Indent      int  `mapstructure:"indent"`
-	LockFile    bool `mapstructure:"lockfile"`
-	Required    bool `mapstructure:"required"`
-	Sensitive   bool `mapstructure:"sensitive"`
-	Type        bool `mapstructure:"type"`
+	Anchor       bool `mapstructure:"anchor"`
+	Color        bool `mapstructure:"color"`
+	Default      bool `mapstructure:"default"`
+	Description  bool `mapstructure:"description"`
+	Escape       bool `mapstructure:"escape"`
+	HideEmpty    bool `mapstructure:"hide-empty"`
+	HTML         bool `mapstructure:"html"`
+	Indent       int  `mapstructure:"indent"`
+	LockFile     bool `mapstructure:"lockfile"`
+	ReadComments bool `mapstructure:"read-comments"`
+	Required     bool `mapstructure:"required"`
+	Sensitive    bool `mapstructure:"sensitive"`
+	Type         bool `mapstructure:"type"`
 }
 
 func defaultSettings() settings {
 	return settings{
-		Anchor:      true,
-		Color:       true,
-		Default:     true,
-		Description: false,
-		Escape:      true,
-		HideEmpty:   false,
-		HTML:        true,
-		Indent:      2,
-		LockFile:    true,
-		Required:    true,
-		Sensitive:   true,
-		Type:        true,
+		Anchor:       true,
+		Color:        true,
+		Default:      true,
+		Description:  false,
+		Escape:       true,
+		HideEmpty:    false,
+		HTML:         true,
+		Indent:       2,
+		LockFile:     true,
+		ReadComments: true,
+		Required:     true,
+		Sensitive:    true,
+		Type:         true,
 	}
 }
 
@@ -510,6 +513,9 @@ func (c *Config) extract() (*print.Settings, *terraform.Options) {
 	options.SortBy.Name = c.Sort.Enabled && c.Sort.By == sortName
 	options.SortBy.Required = c.Sort.Enabled && c.Sort.By == sortRequired
 	options.SortBy.Type = c.Sort.Enabled && c.Sort.By == sortType
+
+	// read comments
+	options.ReadComments = c.Settings.ReadComments
 
 	// settings
 	settings.EscapeCharacters = c.Settings.Escape
--- a/internal/terraform/module_test.go
+++ b/internal/terraform/module_test.go
@@ -498,7 +498,8 @@ func TestLoadInputs(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			assert := assert.New(t)
 			module, _ := loadModule(filepath.Join("testdata", tt.path))
-			inputs, requireds, optionals := loadInputs(module)
+			options := NewOptions()
+			inputs, requireds, optionals := loadInputs(module, options)
 
 			assert.Equal(tt.expected.inputs, len(inputs))
 			assert.Equal(tt.expected.requireds, len(requireds))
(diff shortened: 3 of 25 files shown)


poojitha-bikki:
@khos2ow based on our discussion, can you please review.

poojitha-bikki:
I run `make docs` and it resulted in these changes in commit [26b860b3](https://github.com/terraform-docs/terraform-docs/pull/565/commits/26b860b31e05a1f237cb6be56286828d9895bef1)

When default is option is set to `true` [here](https://github.com/terraform-docs/terraform-docs/pull/565/files#diff-5677b4ba310460bbe02aeef47af1bdb1a8693af0eaa1c46f14f02eb20d65e4b8R53) are these changes expected ?

khos2ow:
> When default is option is set to true here technically there shouldn't be any changes to the docs right ?

You are absolutely correct on this, there shouldn't be any changes in `docs/references/` folder. This clearly indicates there's some flaw in the changes in this PR.

khos2ow:
OK, found the issue. It actually doesn't have anything to do with your implementation, the docs generator is faulty. Can you please add the same property (i.e. `VarDescriptionFromComments` or whatever you renamed it to) to https://github.com/terraform-docs/terraform-docs/blob/master/scripts/docs/generate.go#L177 ?

poojitha-bikki:
> https://github.com/terraform-docs/terraform-docs/blob/master/scripts/docs/generate.go#L177

That's exactly where I was looking. Assumed that for options that are not explicitly set, it'll use default. Got it now. Will update. Thanks!

internal/terraform/options.go
```diff
@@ -26,29 +26,31 @@ type SortBy struct {
...
+	Path                       string
+	ShowHeader                 bool
+	HeaderFromFile             string
+	ShowFooter                 bool
+	FooterFromFile             string
+	UseLockFile                bool
+	SortBy                     *SortBy
+	OutputValues               bool
+	OutputValuesPath           string
+	VarDescriptionFromComments bool
```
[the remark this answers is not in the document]
poojitha-bikki: addressed

khos2ow:
> Assumed that for options that are not explicitly set, it'll use default.

I'm actually in the process of refactoring `Options` which will solve this. It will be as you mentioned, anything now explicitly set in `generate.go` will take the default value.

poojitha-bikki:
@khos2ow added config `ReadComments` (cli flag `read-comments`) for now. Please review. Let me know if the added config setting should be propagated into `printSettings` and `printSDKsettings`

We discussed there might be some overlap with the `description` setting for hcl. But appears `description` solves a different purpose, so my guess is new flag added shouldn't be an issue. Below is an example when both flags are used. 

```
▶ go run main.go tfvars hcl --description --read-comments=true <sample-tf-module>

# some comment
var1 = ""
```
```
▶ go run main.go tfvars hcl --read-comments=true <sample-tf-module>
var1 = ""
```

cmd/root.go
```diff
@@ -81,6 +81,8 @@ func NewCommand() *cobra.Command {
 	cmd.PersistentFlags().BoolVar(&config.OutputValues.Enabled, "output-values", false, "inject output values into outputs (default false)")
 	cmd.PersistentFlags().StringVar(&config.OutputValues.From, "output-values-from", "", "inject output values from file into outputs (default \"\")")
 
+	cmd.PersistentFlags().BoolVar(&config.Settings.ReadComments, "read-comments", true, "use comments for variable description when description is empty")
```
khos2ow, line 84: Use of `variable` here is a bit misleading, comments will be read from all the resources, i.e. variable, output, module, etc.

```suggestion
	cmd.PersistentFlags().BoolVar(&config.Settings.ReadComments, "read-comments", true, "use comments as description when description is empty")
```

internal/cli/config.go
```diff
@@ -40,15 +40,16 @@ var flagMappings = map[string]string{
...
+	"anchor":        "settings.anchor",
+	"color":         "settings.color",
+	"default":       "settings.default",
+	"description":   "settings.description",
+	"escape":        "settings.escape",
+	"indent":        "settings.indent",
+	"required":      "settings.required",
+	"sensitive":     "settings.sensitive",
+	"type":          "settings.type",
+	"read-comments": "settings.read-comments",
```
khos2ow, line 52: Keep this in alphabetic order please.

internal/cli/config.go
```diff
@@ -376,34 +377,36 @@ func (s *sort) validate() error {
...
+	Description  bool `mapstructure:"description"`
+	Escape       bool `mapstructure:"escape"`
+	HideEmpty    bool `mapstructure:"hide-empty"`
+	HTML         bool `mapstructure:"html"`
+	Indent       int  `mapstructure:"indent"`
+	LockFile     bool `mapstructure:"lockfile"`
+	Required     bool `mapstructure:"required"`
+	Sensitive    bool `mapstructure:"sensitive"`
+	Type         bool `mapstructure:"type"`
+	ReadComments bool `mapstructure:"read-comments"`
```
khos2ow, line 392: Keep this in alphabetic order please.

internal/cli/config.go
```diff
@@ -376,34 +377,36 @@ func (s *sort) validate() error {
...
+		Description:  false,
+		Escape:       true,
+		HideEmpty:    false,
+		HTML:         true,
+		Indent:       2,
+		LockFile:     true,
+		Required:     true,
+		Sensitive:    true,
+		Type:         true,
+		ReadComments: true,
```
khos2ow, line 409: Keep this in alphabetic order please.

khos2ow (commented):
Please look at my comments below. Also please update the new config in:

- `README.md`
- `docs/user-guide/configuration.md`
- `docs/user-guide/configuration/settings.md`

khos2ow:
> Let me know if the added config setting should be propagated into printSettings and printSDKsettings

This should be good as is and it's not necessary as this is an entirely internal config to load the module.

khos2ow:
I realized in no point there's a test with `ReadComments: false`. It might be a good idea to create a test case, `TestReadComments()` in `module_test.go`, and for simplicity you can create `internal/terraform/testdata/read-comments` with one `variable` and one `output` with comments and test-cases of `ReadComments` both `true` and `false`.

cmd/root.go
poojitha-bikki: I can be wrong but my understanding is description is only applicable to input variables and output values. If so, does this look ok ?
```suggestion
	cmd.PersistentFlags().BoolVar(&config.Settings.ReadComments, "read-comments", true, "use comments for description of inputs, outputs when description is empty")
```

poojitha-bikki:
- [ ] Updated docs (added new config setting)
- [ ] Added `TestReadComments()` with testcases and testdata to test new config `read-comments`

cmd/root.go
khos2ow: You are correct, it was my bad not to clarify it more. As part of #550 we are going to expand it to all resources; i.e. modules, and resource as well.

poojitha-bikki: I see, understood. took a look at 550, appears we probably need a different flag for that. Users might want to only use comments for description of (inputs , outputs) and not extract any other comments right. Either case, I can commit your suggestion above. we probably don't have to decide right now.

khos2ow (approved):
Thank you @poojitha-bikki for all the efforts. lgtm!
would you please now squash your commits into one?

internal/terraform/module_test.go
```diff
@@ -782,6 +784,49 @@ func TestLoadComments(t *testing.T) {
...
+			fileName:     "variables.tf",
+			readComments: true,
+			expected:     "B description",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert := assert.New(t)
+			options := NewOptions()
+			options.ReadComments = tt.readComments
```
poojitha-bikki, line 814: Just one more question @khos2ow. for me, it took a while to figure out that a `true` boolean flag cannot be overriden (to  `false`) with `options.With` and `.WithOverwrite`. [This comment]((https://github.com/terraform-docs/terraform-docs/blob/master/internal/terraform/module_test.go#L43) helped. do you think it'll be helpful to add it here as well ?  I did not want to add an extra comment if it's not needed.

khos2ow: That's a very good observation, but you can leave it out for now as `Options` and `Settings` will be deprecated in favor of `Config` as part of #510.

poojitha-bikki:
Thank you @khos2ow for your guidance.

PR #616, opened 2022-05-21
fix: sort required format

Rishang:
Signed-off-by: Rishang <rishangbhavsar22@gmail.com>


khos2ow:
The example you modified in this PR belongs to terraform-docs `< v0.13.0` which is intended to be a list.

> As of `v0.13.0`, `sort.by` is converted from `list` to `string`.

PR #657, opened 2022-12-01
Fixed recursive flag stat error

### LINKED ISSUE #654: Flag "--recursive" fails if "modules" folder does not exists
### Describe the bug

When running terraform-docs with --recursive flag in a project without any "modules" folder it fails with "Error: stat modules: no such file or directory".  

### How can we reproduce it?

*command:*

`./terraform-docs markdown table --recursive --output-file Readme.md .` 

*result:*

`Error: stat modules: no such file or directory`

*expected result:*
Readme.md fully rendered

### Environment information
version: 0.16.0, master branch
go version: go1.19.3 linux/amd64
OS: MacOS, Linux

alexyz79:

Changed the behaviour of 'findSubmodules' in 'internal/cli/run.go', by returning an empty module list if the folder is not found, this will allow the normal execution of terraform-docs using '--recursive' even if the 'modules' folder is not available. It fixes when the recursive flag is used on a project without a "modules" folder. 

It only returns an empty list if  'err' is a 'NotExist' error, other wise it keeps the same behaviour: returns the error code.

Fixes #654 

I have:

Tested in terraform projects with and without "modules" folder.

### DIFF
--- a/internal/cli/run.go
+++ b/internal/cli/run.go
@@ -259,7 +259,12 @@ func (r *Runtime) mergeConfig(v *viper.Viper) (*print.Config, error) {
 func (r *Runtime) findSubmodules() ([]module, error) {
 	dir := filepath.Join(r.rootDir, r.config.Recursive.Path)
 
-	if _, err := os.Stat(dir); os.IsNotExist(err) {
+	if _, err := os.Stat(dir); err != nil {
+
+		if os.IsNotExist(err) {
+			return []module{}, nil
+		}
+
 		return nil, err
 	}
 


trenslow:
any updates on getting this merged? 

@khos2ow

khos2ow:
The fix looks good and simple enough, but I don't see the benefit of it. This error, if happens, indicates either of two things: 1) you do intend for the modules folder to be there, but it's not (missing, typo in folder name, etc) 2) you do not intend for the recursive modules to be there, hence the `.terraform-docs.yml` or the command line execution flags are not supposed to have the recursive configs.

In both cases the error seems to be needed, IMO.

cyuanli:
https://github.com/terraform-docs/terraform-docs/issues/654 describes a case where this is needed which I'm experiencing also: A common pipeline for different modules, some of them containing a modules directory but not all of them. I would prefer an option to disable the error, something like `ignore-missing-submodules` or `ignore-missing` or similar.

PR #681, opened 2023-03-22
Render a data or resource block's "Description" attribute in the markdown table template

### LINKED ISSUE #688: Add option diff that when used along with output-check shows differences
Often times it happens that local and automation runs of terraform-docs produce slightly different versions of markdown due to various reasons and if the automation checks for up-to-date versions of markdown files with --output-check, it only shows whether the file is up-to-date or not without showing the actual diff between the produced file and what is in scm. Debugging this becomes harder as you need to replicate the exact environment of the automation and run it to see the difference. 

If there was a diff option ex: `terraform-docs --output-check -diff .` that actually showed you the differences it would become much easier to pin-point why there is a difference in generated markdown v/s what is committed in scm

deepbrook:
Adds a "Comment" column to markdown tables for resources and data blocks.

Fixes #688

### DIFF
--- a/format/templates/markdown_table_resources.tmpl
+++ b/format/templates/markdown_table_resources.tmpl
@@ -8,14 +8,14 @@
     {{ else }}
         {{- indent 0 "#" }} Resources
 
-        | Name | Type |
-        |------|------|
+        | Name | Type | Comment |
+        |------|------|---------|
         {{- range .Module.Resources }}
             {{- $isResource := and $.Config.Sections.Resources ( eq "resource" (printf "%s" .GetMode)) }}
             {{- $isDataResource := and $.Config.Sections.DataSources ( eq "data source" (printf "%s" .GetMode)) }}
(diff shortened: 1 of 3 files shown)


format/templates/markdown_table_resources.tmpl
```diff
@@ -8,15 +8,15 @@
     {{ else }}
         {{- indent 0 "#" }} Resources
 
-        | Name | Type |
-        |------|------|
+        | Name | Type | Comment |
```
jason-johnson, line 11: Why Comment instead of Description?

format/templates/markdown_table_resources.tmpl
```diff
@@ -8,15 +8,15 @@
...
+        |------|------|---------|
```
jason-johnson, line 12: This column should be optional and turned on by a flag.  Not everyone may want this and it would change existing behaviour.

PR #690, opened 2023-06-29
Add fish shell completion

yutaro1985:

- Add completion of [fish shell](https://fishshell.com/)
  - Cobra supports bash, zsh and fish, PowerShell
    - I don't have windows environment, so I only add fish completion.

I have:

I actually built the source and verified that the completion appears on the command line on the fish shell.

### DIFF
--- a/cmd/completion/fish/fish.go
+++ b/cmd/completion/fish/fish.go
@@ -21,7 +21,7 @@ func NewCommand() *cobra.Command {
 	cmd := &cobra.Command{
 		Args:  cobra.NoArgs,
 		Use:   "fish",
-		Short: "Generate shel completion for fish",
+		Short: "Generate shell completion for fish",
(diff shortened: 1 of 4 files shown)


cmd/completion/fish/fish.go
```diff
@@ -0,0 +1,30 @@
...
+
+	"github.com/spf13/cobra"
+)
+
+// NewCommand returns a new cobra.Command for 'completion fish' command
+func NewCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Args:  cobra.NoArgs,
+		Use:   "fish",
+		Short: "Generate shel completion for fish",
```
khos2ow, line 24:
```suggestion
		Short: "Generate shell completion for fish",
```

khos2ow (commented):
Thank you @yutaro1985, it looks good except one small typo here.

Issue #605, opened 2022-03-29
Include Examples supports line selection


aidanmelen:
### What problem are you facing?

https://terraform-docs.io/how-to/include-examples/

Sometimes the `example/main.tf` file has a lot of boilerplate used to setup a module/resource declaration. This is required for the code, but often to verbose for a snippet for a README document.

Consider the [aws instance resource](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/instance). The example in the documentation looks something like this:

```hcl
data "aws_ami" "ubuntu" {
  most_recent = true

  filter {
    name   = "name"
    values = ["ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-*"]
  }

  filter {
    name   = "virtualization-type"
    values = ["hvm"]
  }

  owners = ["099720109477"] # Canonical
}

resource "aws_instance" "web" {
  ami           = data.aws_ami.ubuntu.id
  instance_type = "t3.micro"

  tags = {
    Name = "HelloWorld"
  }
}
```

This could be trimmed down for the sake of brevity

```hcl
resource "aws_instance" "web" {
  ami           = data.aws_ami.ubuntu.id
  instance_type = "t3.micro"

  tags = {
    Name = "HelloWorld"
  }
}
```

The README would remain specific whilst the `example/main.tf` code would maintain the complete set of boilerplate terraform.

### How could terraform-docs help solve your problem?

It would be great if the include examples functionality could select a subset of lines from an example file. This might work similar to Github's [select multiple lines of code feature](https://blog.charlie.fish/posts/2018/10/16/github-select-multiple-lines-of-code/)

The `.terraform-docs.yml` could look something like this.

```yml
# .terraform-docs.yml
content: |-
  {{ .Header }}

  ## Example

  ```hcl
  {{ include "examples/example-1/main.tf#L10-L50" }}
  ```

where `L10` is the first line and `L50` is the last line to include.

  {{ .Inputs }}

  {{ .Outputs }}  
```

### Alternative
You could refactor the example such that the `main.tf` only contains the snippet you would like to include in the README, but then we are forced to format our examples in unfavorable ways to make the docs work.

Issue #599, opened 2022-03-10
Optional include statement


thomasfinstad:
### What problem are you facing?

Includes causes the execution to fail if the file does not exist.
Forcing either empty files where they are not needed, or custom templates for each situation.

This is especially annoying in a recursive setup with submodules where the template is only defined at the top level in .terraform-docs.yaml

### How could terraform-docs help solve your problem?

An include statement that checks for the file first and skips it if it is missing.

Suggestions for syntax: 

```
{{ optional_include "examples/example-1/main.tf" }}
```

```
{{ include_nofail "examples/example-1/main.tf" }}
```

```
{{ include "?examples/example-1/main.tf" }}
```

Joerg-L:
Yeah that would be really a good extension.
We are hitting the same issue

- shared config file for terraform-docs 
- modules with and without examples

mgzenitech:
Either that or just additional argument to return that value instead of file content:
```
{{ include "somerandomfile" "" }}
```

In this case it would return empty string instead of panic

seniorheuser:
We also need this.

PR #709, opened 2023-09-16
Add an Examples section to automatically pull in module examples

### LINKED ISSUE #617: auto-populate a {{ .Examples }} from the examples directory

I would like to use an `./examples` directory as prescribed by [the standard module structure](https://www.terraform.io/language/modules/develop/structure).

The documentation suggests using `include`:

```
  {{ include "examples/example1/main.tf" }}
```

This definitely works and is still pretty easy. However, since the examples directory is a standard, it would be cool to have that part automatically generated as well, and made available to `content` via `{{ .Examples }}`.

Automatically recognizing the `/examples` directory and using its contents for an "Examples" section of the documentation.

### LINKED ISSUE #685: Implement an {{ .Examples }} content block

I want a better way to represent Examples in terraform-docs. The [current method](https://terraform-docs.io/how-to/include-examples/) works however it can be a bit troublesome when scaling out. My particular use case is that I have many Terraform module repositories in a project within my company. When a new version is pushed it will automatically run terraform-docs to generate the documentation. If I want to add an Examples section to it, I must either ensure that every repo has an example in place or I would need to add some kind of logic to inject in the `##Examples` string only if the file exists. This gets more complicated for multiple examples. I'm hoping to make a easier, more native way for people to include examples. 

I propose adding a `{{ .Header }}` content block that will have the default behaviour of iterating through `examples/` and writing blocks for each main.tf found in child folders of `examples` with the subtitle `## Examples`
E.g.
````
## Examples
```hcl
{{ include "examples/exampleA/main.tf" }}
```
```hcl
{{ include "examples/exampleB/main.tf" }}
```
````
*This may or may not use the `include`  function, this is just a representation of what it would do*

If none of these files are found, no blocks would be created. Ideally the title will also be removed, but could also settle with "No Examples were found for this module". 

Possible options for the block could be: 
- Include / exclude options in case people want specific examples shown 
- Come kind of filtering option for the examples (e.g. only taking certain lines numbers)

I'm hoping to do the legwork on this issue but I wanted to follow the guidelines to see if this is a desired change / if this is already underway at all. 

Thanks :)

Credit to these issues which overlap / gave inspiration:
#617
#605 
#599

### PR DESCRIPTION
**Pull request by Varantha**
#### Description 

Adding an `Examples` section into the tool. The purpose being that if you have a list of examples in a subfolder in your Terraform module repository it will automatically find them and insert them into your documentation. 

For example, I've now added an example to the example module here:
[examples/examples/exampleA/main.tf](examples/examples/exampleA/main.tf)

This takes the modules name (folder name) exampleA and its content (the file contents) and will display it as an additional section if configured to do so. It will take all folder names from 1 level below the `examples` folder, so long as that folder has a `main.tf` file present.

#### Added Config 

Some extra config has been added to support this change. 

`examples-from`: Following the style of footer-from and header-from, the location to take the examples from. This should point to the folder containing module folders. The default value is `./examples` (aligning with Terraform's [Standard Module Structure](https://developer.hashicorp.com/terraform/language/modules/develop/structure))

`section` > `hide`/`show`: Adding `examples` to the hide and show functionality. Example's default state is being hidden to avoid this change being disruptive

`examples` > `include`: A list of strings for what examples to include. These strings match the names of the example, in this case you would put: 

```yml
include:
    - exampleA
```
To include the example in this project. All examples will be shown if neither include or exclude are set.

`examples` > `exclude`: A list of strings for what examples to exclude. This is the same functionality as `include` but in reverse. (attempting to match `show` / `hide`)

`examples` > `limit`: An integer for the maximum number of examples to show. If a project has too many examples then you might want to limit them and at scale you may not be able to include specific examples by name. This takes the first x examples in lexicographic order. If the value is set to 0 then all examples will be shown. The default is 0. 

`examples` > `HideTitles`: By default, in formats like markdown and ascii it will show the title of the example as a header above it. e.g.

```
## Examples

### exampleA

```hcl
module full-example{
    source = "github.com/terraform-docs/terraform-docs//examples"
...
```

but this can be set to `true` to hide the title such that it is just: 
```
## Examples

```hcl
module full-example{
    source = "github.com/terraform-docs/terraform-docs//examples"
...
```

#### Documentation

I'm very happy to update the documentation along with this change but I wanted to gauge interest in the feature / the config setting before continuing with it. (as it may change) Please let me know your thoughts on this and I'll start work on it. 

Fixes #685
Fixes #617 (Not mine, but confident it's the same purpose)
Possibly #599 / #605 (Not my issues and unsure if it 100% meets their needs)
I have:

I've added a whole suite of tests /updated tests in for every type of output which support `examples` and am confident that it has functionality in those areas. I've installed this branch locally and have applied it to some of my local projects without issue and have been messing around with the configuration file / combinations of configurations. Happy for feedback on any tests that could be added / edge cases you would like testing, but it seems good for standard use. 

I would be interested for feedback for the more unique formats. I've updated the tests and tried to match the general style of format like pretty / toml / json, but these are typically not my use-cases so I can't say that they are ideally formatted in those areas.

### DIFF
--- a/examples/examples/exampleA/main.tf
+++ b/examples/examples/exampleA/main.tf
@@ -1,0 +1,17 @@
+module full-example{
+    source = "github.com/terraform-docs/terraform-docs//examples"
+
+    input_with_underscores = "input_any"
+    list-2 = [
+        "value_1",
+        "value_2"
+    ]
+    map-2 = {
+        "1" = "value"
+        "2" = "value_2"
+    }
+    number-2 = 2
+    string-2 = "2"
+    string_no_default = "string_value"
+    unquoted = test
+}
--- a/examples/.terraform-docs.yml
+++ b/examples/.terraform-docs.yml
@@ -19,6 +19,7 @@ footer-from: footer.md
 sections:
   show:
     - header
+    - examples
     - inputs
     - providers
     - modules
@@ -87,5 +88,5 @@ settings:
   indent: 4
   escape: false
   default: false
-  required: false
+  required: true
   type: true
--- a/format/asciidoc_document_test.go
+++ b/format/asciidoc_document_test.go
@@ -106,6 +106,11 @@ func TestAsciidocDocument(t *testing.T) {
 		"OnlyDataSources": {
 			config: testutil.With(func(c *print.Config) { c.Sections.DataSources = true }),
 		},
+		"OnlyExamples": {
+			config: testutil.With(func(c *print.Config) {
+				c.Sections.Examples = true
+			}),
+		},
 		"OnlyHeader": {
 			config: testutil.With(func(c *print.Config) { c.Sections.Header = true }),
 		},
--- a/format/asciidoc_table_test.go
+++ b/format/asciidoc_table_test.go
@@ -106,6 +106,11 @@ func TestAsciidocTable(t *testing.T) {
 		"OnlyDataSources": {
 			config: testutil.With(func(c *print.Config) { c.Sections.DataSources = true }),
 		},
+		"OnlyExamples": {
+			config: testutil.With(func(c *print.Config) {
+				c.Sections.Examples = true
+			}),
+		},
 		"OnlyHeader": {
 			config: testutil.With(func(c *print.Config) { c.Sections.Header = true }),
 		},
--- a/format/generator.go
+++ b/format/generator.go
@@ -87,6 +87,13 @@ func withResources(resources string) generateFunc {
 	}
 }
 
+// withExamples specifies how the generator should add Examples.
+func withExamples(examples string) generateFunc {
+	return func(g *generator) {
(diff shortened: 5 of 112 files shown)


rquadling:
For more complex modules, examples are a VERY useful feature. Just as within a provider, not all properties are needed if one of the properties has a value, so, rather than JUST documenting that in each variable, worked examples, which may end up being copy'n'pasted by some, is a clear outline of the expected use of a module.

examples/examples/exampleA/main.tf
```diff
@@ -0,0 +1,17 @@
+module full-example{
+    source = "github.com/terraform-docs/terraform-docs//examples"
```
rquadling, line 2: The double `//`? Probably fine, but something different always leads to the question "why".

Issue #887, opened 2025-12-17
0.21.0 not available as docker image


antdking:
### Describe the bug

```
% date
Wed Dec 17 19:07:45 GMT 2025
% docker pull quay.io/terraform-docs/terraform-docs:0.21.0
Error response from daemon: manifest for quay.io/terraform-docs/terraform-docs:0.21.0 not found: manifest unknown: manifest unknown
```

similar to this issue: https://github.com/terraform-docs/terraform-docs/issues/846

Looking in Quay, both `0.21.0-amd64` and `0.21.0-arm64` exist.

RafaelWO:
Note that the README still mentions `quay.io/terraform-docs/terraform-docs:0.21.0`. So this looks like the new tags with the architecture suffix are not intentional, right? 👀

ElvenSpellmaker:
Same applies to `0.22.0` too.

RafaelWO:
This is still an issue – would be great if the arch suffix in the Docker tags could be removed. 

<details><summary>Screenshot</summary>

<img width="1506" height="1102" alt="Image" src="https://github.com/user-attachments/assets/baa7cf0c-9dc6-4228-9665-3d2fca8c851b" />

</details> 

It looks like there are already some candidate PRs open. Thanks in advance for getting one of them merged 🙂

ShuP1:
The wrong tag is still mentioned in https://github.com/terraform-docs/terraform-docs/releases 😢 

> docker pull quay.io/terraform-docs/terraform-docs:0.24.0

I agree than a multi-arch semantic versioning tag would be great

PR #894, opened 2026-01-10
docs: clarify container image registry

Piyushkhobragade:
Clarifies that terraform-docs Docker images are published on Quay and not Docker Hub,
to avoid confusion when pulling tagged releases.

Related to #887

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,8 +84,12 @@ Check [`output`] configuration for more details and examples.
 
 ### Using docker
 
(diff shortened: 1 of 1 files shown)


pureiboi:
i think the docker container is specified in the installation instruction, might not make sense to add another place?

https://terraform-docs.io/user-guide/installation/#docker

https://github.com/terraform-docs/terraform-docs/blob/master/docs/user-guide/installation.md#docker

PR #897, opened 2026-02-02
Open tofu dynamic providers

### LINKED ISSUE #895: OpenTofu dynamic providers break execution
Since upgrading to 0.20.0 we can no longer run it on existing project that has a mixture of `.tf` and `.tofu`. We'd accepted that `.tofu` wasn't included in the documentation.

We're using dynamic providers in tofu files which are causing the failures:

```shell
Error: Invalid provider reference: Provider argument requires a provider name followed by an optional alias, like "aws.foo". (and 4 other messages)
```

### Describe the bug

Adding the `.tofu` files in #833 has broken the ability to generate documents on projects with a mix of files.

### How can we reproduce it?

```terraform
# main.tofu
locals {
  regions = toset(["us-east-1", "eu-west-1"])
}

provider "aws" {
  for_each = local.regions
  alias    = "main"
  region   = each.key
}

data "aws_caller_identity" "current" {
  for_each = local.regions
  provider = aws.main[each.key]
}

output "aws_caller_identity_arn" {
  value = values(data.aws_caller_identity.current)[*].arn
}
```

### Environment information
- terraform-docs version (use `terraform-docs --version`): 0.21.0

### Additional

We've downgraded to 0.19.0.

guanana:
I was reading about the issue and I was affected by this myself. This PR tries to solve the "delicate" problem of OpenTofu for_each provider creating a fatal error in the parsing. Maybe not the best solution but I hope that a solution nevertheless. I'm not too good with **go** so I helped myself with an "assistant" to help me complete the task.

I'm aware that this fix maybe doesn't belong here but it would allow a lot of projects (like mine) to continue using your great tool terraform-docs even when using dynamic providers with OpenTofu

Added logic to loadModule to ignore specific "Invalid provider reference" errors that occur during the initial HCL load.
Provider Fix (fixOpenTofuProviders): 
Implemented a function that:

- Iterates through resources in the loaded module.
- If a resource is missing its provider name (due to the parsing error), it re-parses the source file locally.
- It correctly interprets the for_each index expression (e.g., aws.main[each.key]) to extract the provider name and alias.
- Updates the resource in the module with the correct provider information.

Fixes #895 
I have:

- [X] Read and followed terraform-docs' [contribution process].
- [X] All tests pass when I run `make test`.

Added a new test case TestLoadOpenTofuProviders in terraform/load_test.go and corresponding test data in terraform/testdata/opentofu-for-each/.

### DIFF
--- a/terraform/load.go
+++ b/terraform/load.go
@@ -22,7 +22,10 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/hashicorp/hcl/v2"
+	"github.com/hashicorp/hcl/v2/hclparse"
 	"github.com/hashicorp/hcl/v2/hclsimple"
+	"github.com/hashicorp/hcl/v2/hclsyntax"
 
 	"github.com/terraform-docs/terraform-config-inspect/tfconfig"
 	"github.com/terraform-docs/terraform-docs/internal/reader"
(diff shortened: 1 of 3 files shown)


voidtek:
Hi

I just tested this branch on our project which uses OpenTofu dynamic providers and it works perfectly — no more Invalid provider reference error.

Is there any timeline for when this will be merged and released? We'd love to move off our workaround. 

Thanks for the fix!

vitotafuni:
up

PR #905, opened 2026-04-07
feat: make registry URL configurable for OpenTofu and private registries

### LINKED ISSUE #812: Make the registry URL configurable, and provider/module version more accurate.

Presently, the registry URL is hard-coded to `registry.terraform.io`.

With the advent of OpenTofu, more registries have popped up, including better support for internal/enterprise/corporate "private" registries. A user of this tool should be able to specify the preferred URL pattern to use for the registry.

The three most popular _public_ registries (as of today's date) have the following URL patterns (arbitrarily picking version `v5.65.0`).

* https://registry.terraform.io/providers/hashicorp/aws/5.65.0/docs
* https://search.opentofu.org/provider/hashicorp/aws/v5.65.0
* https://library.tf/providers/hashicorp/aws/5.65.0

I propose that as a CLI flag and config file entry, a user should be able to provide a keyword to signify one of the popular public registries, or a Go-template URL pattern for a custom/private registry.

As an example, let's look at [settings](https://terraform-docs.io/user-guide/configuration/settings/). We could add a new entry called `registry-url`, which would accept a string.

```yaml
settings:
  anchor: true
  color: true
  default: true
  registry-url: ''
  ...
```

Reviewing [When not to use Plain Scalars](https://www.yaml.info/learn/quote.html#noplain) to understand which characters can be used at the beginning of a string value, unquoted, it looks like `:<space>` is unavailable, but `:<nospace>` should be valid.

We could use certain "short" strings as placeholders for well-known values, then fall back to Go-template values for custom URLs.

* `:terraform:`
* `:opentofu:`
* `:library-tf:`
* `https://registry.company.com/internal-registry/{{.Namespace}}/{{.Provider}}/{{.ProviderVersion}}/docs`

The last of these would likely _need_ to be quoted, which is reasonable.

```yaml
settings:
  default: true
  registry-url: :library-tf: # I *think* this is valid. Untested. Can still be quoted.
  ...
```

```yaml
settings:
  default: true
  registry-url: 'https://registry.company.com/internal-registry/{{.Namespace}}/{{.Provider}}/{{.ProviderVersion}}/docs'
  ...
```

Or if you'd prefer to simply require a quoted value, maybe a different meta-character would be preferred over `:` for readability. Perhaps `@`?

```yaml
settings:
  default: true
  registry-url: '@opentofu'
  ...
```

Just some ideas.

### Related improvements

If a provider has a version constraint (e.g., `~> 4.60`), then instead of always linking to the `latest` version, we should honor the version constraint when we create the registry links. Continuing to use `hashicorp/aws` as an example, the most appropriate version for the `~> 4.60` constraint is `4.67.0`.

In my proposal above, where I mentioned `.ProviderVersion` as a Go-template variable, that variable would be set to `4.67.0` instead of `latest`.

**Why?** Many corporations have large numbers of apps in various states of maintenance, and may be using an older version of Terraform or Terraform Module or Terraform Provider until they have a chance to update. My previous company has in-use versions of Terraform 0.7 that just never got updated. Being able to provide _accurate_ documentation for well-crafted projects should be possible, and always linking to `latest` is _inaccurate_.

thenoid:
Adds a `registry-url` setting (and `--registry-url` CLI flag) that accepts a Go template string for customizing resource documentation links. Supports template vars: {{.Namespace}}, {{.Provider}}, {{.Version}}, {{.VersionWithV}}, {{.Kind}}, {{.Type}}. Defaults to the Terraform public registry (no behavior change).

Fixes #812

I have:

- [ ] Read and followed terraform-docs' [contribution process].

### DIFF
--- a/terraform/resource.go
+++ b/terraform/resource.go
@@ -11,9 +11,10 @@ the root directory of this source tree.
 package terraform
 
 import (
-	"fmt"
+	"bytes"
 	"sort"
 	"strings"
+	gotemplate "text/template"
 
 	"github.com/terraform-docs/terraform-docs/internal/types"
 )
@@ -28,7 +29,10 @@ type Resource struct {
 	Version        types.String `json:"version" toml:"version" xml:"version" yaml:"version"`
 	Description    types.String `json:"description" toml:"description" xml:"description" yaml:"description"`
 	Position       Position     `json:"-" toml:"-" xml:"-" yaml:"-"`
+	RegistryURL    string       `json:"-" toml:"-" xml:"-" yaml:"-"`
 }
+
+const defaultRegistryURLTemplate = "https://registry.terraform.io/providers/{{.Namespace}}/{{.Provider}}/{{.Version}}/docs/{{.Kind}}/{{.Type}}"
 
 // Spec returns the resource spec addresses a specific resource in the config.
 // It takes the form: resource_type.resource_name[resource index]
@@ -50,7 +54,8 @@ func (r *Resource) GetMode() string {
 	}
 }
 
-// URL returns a best guess at the URL for resource documentation
+// URL returns the URL for resource documentation using the configured registry URL template.
+// If no custom template is set, it defaults to the Terraform public registry.
 func (r *Resource) URL() string {
 	kind := ""
 	switch r.Mode {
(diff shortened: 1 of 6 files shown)


terraform/resource.go
```diff
@@ -65,7 +70,51 @@ func (r *Resource) URL() string {
...
+	if version != "latest" && version != "" {
+		versionWithV = "v" + version
+	}
+
+	urlTemplate := r.RegistryURL
+	if urlTemplate == "" {
+		urlTemplate = defaultRegistryURLTemplate
+	}
+
+	tpl, err := gotemplate.New("registry-url").Parse(urlTemplate)
```
tmccombs, line 92: Could the template be parsed once, and re-used instead of parsing it every time?

Issue #622, opened 2022-07-12
Disabling sort providers order inconsistent


mcwarman:
### Describe the bug

When disabling sort (so the inputs are in order of the `variables.tf`), the provider orders are inconsistent from execution to execution. 

Flipping between:

```text
| Name | Version |
|------|---------|
| <a name="provider_google"></a> [google](#provider\_google) | >= 4.5.0 |
| <a name="provider_google-beta"></a> [google-beta](#provider\_google-beta) | >= 4.5.0 |
```

```text
| Name | Version |
|------|---------|
| <a name="provider_google-beta"></a> [google-beta](#provider\_google-beta) | >= 4.5.0 |
| <a name="provider_google"></a> [google](#provider\_google) | >= 4.5.0 |
```

### How can we reproduce it?

```hcl
#main.tf
terraform {
  required_providers {
    google = {
      source  = "hashicorp/google"
      version = ">= 4.5.0"
    }
    google-beta = {
      source  = "hashicorp/google-beta"
      version = ">= 4.5.0"
    }
  }
}

resource "google_container_node_pool" "node_pool_1" {
}

resource "google_container_node_pool" "node_pool_2" {
  provider = google-beta
}

resource "google_container_node_pool" "node_pool_3" {
}
```

Run the following repeatably observing output:

```shell
terraform-docs markdown table . --sort=false
```

Alternatively:

```shell
docker run --rm --volume "$(PWD):/terraform-docs" --workdir /terraform-docs quay.io/terraform-docs/terraform-docs:0.16.0 markdown table . --sort=false
```

### Environment information

Tested with both docker (`quay.io/terraform-docs/terraform-docs:0.16.0`) and windows:

```shell
terraform-docs --version
terraform-docs version v0.16.0 1f686b1 windows/amd64
```

acelebanski:
Running terraform-docs v0.20.0 and having the exact same issue. I don't want to sort and the result of my terraform-docs hook is indeterministic, randomly changes the order of providers in random README files across the codebase.

PR #920, opened 2026-04-15
fix: deterministic sorting for providers

### LINKED ISSUE #898: Undeterministic results of terraform-docs execution
### Describe the bug

When I run terraform-docs (as a pre-commit hook in my case), my list of Terraform providers is randomly reordered. I have multiple READMEs in my repository and I never know which ones will be edited by terraform-docs. 

I get this:
```
Terraform docs...........................................................Failed
- hook id: terraform_docs
- files were modified by this hook
```

And that's the result on one or more files:

Image

### How can we reproduce it?

You can copy the code from [this repo](https://github.com/PaloAltoNetworks/terraform-azurerm-swfw-modules).

### Environment information

- terraform-docs version v0.21.0 darwin/arm64
- Go version go1.25.3 darwin/arm64
- OS (e.g. Windows, MacOS): macOS 15.7.3 (24G419)

gusandrioli:

Fixes #898 

Providers and resources were returned in a non-deterministic order due to Go's randomized map iteration. This caused terraform-docs to produce different output on repeated runs, failing pre-commit hooks unnecessarily.
                                                                                                                                                                                
Three fixes in terraform/load.go and terraform/provider.go:                                                                                                                   
   
1. Collect from maps in sorted key order - both loadProviders and loadResources now sort map keys before building the result slice, so the input to all sort functions is always deterministic.                                     
2. Track the earliest position per provider - when a provider is referenced by multiple resources, its position is now always the earliest file+line across all its resources, not whichever resource happened to be iterated first.                                                                                                                        
3. Add a name tiebreaker to sortProvidersByPosition - prevents non-determinism when two providers share the same filename and line number.
                                                                                                                                                                                
- Added testdata/multi-provider-resources together with new unit tests (TestLoadProvidersDeterministic, TestLoadResourcesDeterministic).

I have:

### DIFF
--- a/terraform/load.go
+++ b/terraform/load.go
@@ -402,7 +402,12 @@ func loadProviders(tfmodule *tfconfig.Module, config *print.Config) []*Provider 
 			}
 
 			key := fmt.Sprintf("%s.%s", r.Provider.Name, r.Provider.Alias)
-			if _, ok := discovered[key]; ok {
+			if existing, ok := discovered[key]; ok {
+				// keep the earliest position across all resources of this provider
+				if r.Pos.Filename < existing.Position.Filename ||
+					(r.Pos.Filename == existing.Position.Filename && r.Pos.Line < existing.Position.Line) {
+					existing.Position = Position{Filename: r.Pos.Filename, Line: r.Pos.Line}
+				}
 				continue
 			}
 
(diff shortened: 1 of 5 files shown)


gusandrioli:
As per my comment [on issue #898](https://github.com/terraform-docs/terraform-docs/issues/898#issuecomment-4274513262), this PR also addresses https://github.com/terraform-docs/terraform-docs/issues/622

pascal-hofmann (approved):

Issue #661, opened 2023-01-19
Allow recursive runs to exclude subpaths and/or the top-level path


whoracle:
### What problem are you facing?

We currently have X repositories with Y modules in them (and the modules themselves pull "real" modules via a private module registry in gitlab). I want to  force README generation in each module via pre-commit, but I want to leave the repo root README alone (since there's no tf files in there anyways), but I only want to have one unified terraform-docs.yml in each repository.

Currently, if I run `terraform-docs -c .terraform-docs.yml .` in the repo root, my main README gets modified alongside the module README files. If I start one level deeper with e.g. `terraform-docs -c .terraform-docs.yml subscriptions`, then I get a README in subscriptions, which I also don't want.

Example repo structure:
```
~/terraform/azure
├───subscriptions
│       │
│       ├───my_dev_sub
│       │        ├───main.tf
│       │        ├───more_files.tf
│       │        └───README.md
│       │
│       ├───my_staging_sub
│       │        ├───main.tf
│       │        ├───more_files.tf
│       │        └───README.md
│       │
│       └───my_prod_sub
│                ├───main.tf
│                ├───more_files.tf
│                └───README.md
|
├───.terraform-docs.yml
├───terragrunt.hcl
└───README.md
```

### How could terraform-docs help solve your problem?
I have two ideas for a solution:
- Let me specify which folders to exclude in .terraform-docs.yml in the `recursive` tree or a global .terraform-docs-ignore file in the repo root
- make terraform-docs ignore all directories that contain a .terraform-docs-ignore file in them (can be empty)
For my specific use case I don't have a strong preference either way, since I only want to exclude a single location (the top level), but a more flexible approach would maybe be advised for people who want to ignore more than one path.

Kind Regards

kraduk:
sub optimal, but just .gitignore those files

rob-borg:
As an addition I would like to specify which submodules to exclude from doc generation completely during a recursive generation

martin-reich:
As mentioned in #613: 
You can use a workaround: 

```
terraform-docs markdown --output-mode replace --output-file README.md --recursive --recursive-path ../ ./modules/name-of-submodule
```

nitrocode:
I was looking for a way to ignore specific files because there is a convention in some orgs to add a `variables-deprecated.tf` prior to going to the next major release of a module and it would be nice to exclude this file when autogenerating the docs to reduce the likelihood that someone would use a deprecated variable.

[...]

PR #924, opened 2026-04-19
feat: improve recursive submodule scanning to traverse nested directories and add exclude support

### LINKED ISSUE #585: Recursively find and create docs for all nested modules

For folder structures where the `modules` directory contains nested folders, choosing recursive enabled option currently does not find any additional submodules that are nested in folders. In the example below, it will not detect `nested-example-2` directory.
```
├── modules
│   ├── nested-example-1
│   │   ├── README.md
│   │   ├── main.tf
│   │   ├── nested-example-2
│   │   │   ├── main.tf
│   │   │ 
│   └── .terraform-docs.yml

```

It would be nice to have the recursive enabled option generate docs for folders at the module level that have nested modules in them.

rudransh-shrivastava:
Fixes #585 might also address #661

This PR fixes a lot of created issues but the most relevant is #585
An alternative fix is to allow specifying multiple paths, but it's harder to maintain I think.

I have:

I tested this against our infrastructure code at [OWASP Nest](https://github.com/OWASP/Nest/).

Run with `exclude: tests`
image

Run without `exclude` (affects `tests/`):
image

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -63,6 +63,7 @@ func NewCommand() *cobra.Command {
 	cmd.PersistentFlags().BoolVar(&config.Recursive.Enabled, "recursive", false, "update submodules recursively (default false)")
 	cmd.PersistentFlags().StringVar(&config.Recursive.Path, "recursive-path", "modules", "submodules path to recursively update")
 	cmd.PersistentFlags().BoolVar(&config.Recursive.IncludeMain, "recursive-include-main", true, "include the main module")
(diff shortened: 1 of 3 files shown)


rudransh-shrivastava:
I will add a way to ignore directories like `tests/`

rudransh-shrivastava:
Hi @pascal-hofmann 
could you please review when you get a chance?

pascal-hofmann (approved):

PR #931, opened 2026-05-10
fix: ignore directories without `.tf` files for docs generation

### LINKED ISSUE #929: `recursive`: directories without `.tf` files are treated as modules in v0.23.0+
Hi! Thanks for terraform-docs. We use it across a large infra monorepo and have done so for years.

### Setup

Our layout is the conventional one. Each `modules/<name>/` is one Terraform module:

```
modules/
├── module-a/
│   ├── main.tf, variables.tf, outputs.tf, ...
│   ├── README.md          (terraform-docs injects here)
│   ├── examples/basic/    (sample .tf, not a real module)
│   ├── tests/             (test fixtures, some with .tf)
│   ├── releases/          (release manifests, no .tf)
│   ├── templates/         (helm templates, no .tf)
│   └── scripts/           (shell helpers, no .tf)
├── module-b/
└── ...
```

terraform-docs runs in pre-commit locally and the same hook runs in CI as a determinism check. Any local-vs-CI diff fails the build, so the tool's output behavior matters a lot to us.

### What changed in v0.23.0

#585 was implemented by switching `findSubmodules` from a one-level `os.ReadDir` to a recursive `filepath.WalkDir`. The walker now treats every directory under `recursive.path` as a module candidate, regardless of whether the directory contains any `.tf` files. The result in our repo is 100+ freshly-created READMEs in helper subdirectories (`templates/`, `releases/`, `scripts/`, `tpl/`, `dashboards/`, plus dozens of test-fixture dirs), each containing:

```markdown
<!-- BEGIN_TF_DOCS -->
## Inputs

No inputs.

## Outputs

No outputs.
<!-- END_TF_DOCS -->
```

### Why existing knobs don't help

`recursive.exclude` is basename-only. We would need an exclude list with 100+ entries that grows forever as new helper dirs are added, and basename collisions could accidentally hide a real module. The closest open request is #661, but path-based excludes still require enumerating every non-module dir.

### What we're doing today

Pinning to `~> 0.22.0`, where the one-level `os.ReadDir` matched our convention. We're aware this means we miss future fixes (including the nice deterministic-provider-sort change from #898).

### Could this use case be supported in v0.23+?

Two ideas, either of which would let us upgrade:

1. **Skip dirs that contain no `.tf`/`.tf.json` files.** A dir with no Terraform sources cannot be a module, so generating a README there is never useful. Could be the default, or gated behind `recursive.skip-empty: true` if backward compatibility for #585 is a concern.
2. **A `recursive.depth` option.** Setting it to `1` would restore the v0.22 behavior. The default could stay deep to preserve #585.

Happy to send a PR for option 1 or 2 if the direction is welcome. Wanted to check first. Thanks!

rudransh-shrivastava:
Fixes #929 

My previous PR #924 introduced a bug, I want to fix it.

I have:

Tested this against a local project; A friend also tested this independently.

### DIFF
--- a/internal/cli/run.go
+++ b/internal/cli/run.go
@@ -323,6 +323,27 @@ func (r *Runtime) loadSubModule(path string) (*module, error) {
 	return &module{rootDir: path, config: cfg}, nil
 }
 
+// containsTerraformFiles reports whether a directory contains Terraform files.
+func containsTerraformFiles(path string) (bool, error) {
+	files, err := os.ReadDir(path)
+	if err != nil {
+		return false, err
+	}
+
+	for _, file := range files {
+		if file.IsDir() {
+			continue
+		}
+
+		name := file.Name()
+		if strings.HasSuffix(name, ".tf") || strings.HasSuffix(name, ".tf.json") {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
 // loadModuleConfig attempts to load a module configuration from the given directory path.
 func (r *Runtime) loadModuleConfig(path string) (*print.Config, error) {
 	var cfg *print.Config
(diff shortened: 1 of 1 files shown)


rudransh-shrivastava:
@pascal-hofmann please check whenever you're free!

pascal-hofmann (approved):

pascal-hofmann:
Thank you for the fix! 🚀

PR #951, opened 2026-07-29
Support Terraform 1.15 const variables in module source and version

### LINKED ISSUE #927: Add support for Terraform 1.15

issue:

```sh
╰─ terraform-docs .
Error: Variables not allowed: Variables may not be used here. (and 29 other messages)
```

```sh
─ terraform-docs -v
terraform-docs version v0.22.0 795da6b darwin/arm64
```

Reproduce:

```terraform
variable "version" {
    default = "6.6.1"
    const   = true
    type    = string
}

module "vpc" {
    source  = "terraform-aws-modules/vpc/aws"
    version = var.version  # this

    name = "my-vpc"
    cidr = "10.0.0.0/16"
}
```

Terraform added support for variables in module sources starting for [v1.15](https://github.com/hashicorp/terraform/releases/tag/v1.15.0).

hashicorp/terraform#38160
hashicorp/terraform/pull/38217

Please help us use terraform v1.15 by adopting the changes for `terraform-docs`.

gss2002:

This pull request updates `terraform-docs` to support Terraform 1.15's new capability of using constant variables and locals in module `source` and `version` attributes.

Terraform 1.15 introduced support for expressions such as:

```hcl
variable "version" {
  type    = string
  const   = true
  default = "6.6.1"
}

module "vpc" {
  source  = "terraform-aws-modules/vpc/aws"
  version = var.version
}
```

Currently, `terraform-docs` fails while loading the module with errors similar to:

```
Variables not allowed: Variables may not be used here.
```

The underlying issue is that the bundled `terraform-config-inspect` dependency predates Terraform 1.15 and attempts to decode `source` and `version` as literal strings using an empty evaluation context.

This change updates `terraform-docs` to use the upstream parser implementation that tolerates these expressions by preserving the raw HCL expression instead of returning a fatal diagnostic. This matches Terraform 1.15's language behavior and allows documentation generation to continue without requiring users to hardcode module versions or sources.

Fixes #927

I have:

Testing included the following scenarios:

- Verified that modules using literal `source` and `version` values continue to behave as before.
- Added a regression test covering a Terraform 1.15 module that uses a `const` input variable for the module `version`.
- Verified that module loading no longer fails when `source` or `version` reference `const` variables or local values.
- Confirmed that module call metadata preserves the original expression (for example, `var.version`) rather than producing a fatal parser error, allowing `terraform-docs` to generate documentation successfully.

This change is intentionally limited to parser compatibility and does not attempt to evaluate Terraform expressions. It simply adopts the upstream `terraform-config-inspect` behavior introduced to support Terraform 1.15 while preserving existing functionality for earlier Terraform versions.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,7 @@ require (
 	github.com/spf13/pflag v1.0.10
 	github.com/spf13/viper v1.21.0
 	github.com/stretchr/testify v1.11.1
-	github.com/terraform-docs/terraform-config-inspect v0.0.0-20250408153412-5b88c7ed5b63
+	github.com/hashicorp/terraform-config-inspect v0.0.0-20260709150029-2fb54c236733
 	golang.org/x/exp v0.0.0-20251209150349-8475f28825e9
 	gopkg.in/yaml.v3 v3.0.1
 	honnef.co/go/tools v0.7.0
--- a/terraform/load.go
+++ b/terraform/load.go
@@ -24,7 +24,7 @@ import (
 
 	"github.com/hashicorp/hcl/v2/hclsimple"
 
-	"github.com/terraform-docs/terraform-config-inspect/tfconfig"
+	"github.com/hashicorp/terraform-config-inspect/tfconfig"
 	"github.com/terraform-docs/terraform-docs/internal/reader"
 	"github.com/terraform-docs/terraform-docs/internal/types"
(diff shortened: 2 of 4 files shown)


gss2002:
@pascal-hofmann  any chance we can at least get this into place in the interim so that we can unblock later versions of terraform and terraform-docs? And than look at what things dont work with opentofu?

PR #952, opened 2026-08-01
fix: show "No resources." instead of an empty table when the visible resource kind has no entries

### LINKED ISSUE #603: Hide data-sources doesn't hide them
`terraform-docs version v0.16.0 1f686b1 linux/amd64` 

config file:
```
sections:
  hide:
    - resources
    - data-sources
```

The resources are being hidden but the section remains with data sources.

```
## Resources

| Name | Type |
|------|------|
| [aws_caller_identity.current](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/caller_identity) | data source |
```

ankit090701:
## Summary

`sections.hide` (or `--hide`) for `resources`/`data-sources` doesn't actually hide an empty "Resources" section: a module containing only resources of the currently-hidden kind (e.g. all managed resources, with `sections.hide: [resources]`) still prints an empty `## Resources` table with just column headers, instead of the `No resources.` placeholder every other empty section uses.

```console
$ terraform-docs markdown --hide providers --hide resources .
## Resources

| Name | Type |
| ---- | ---- |

## Inputs
...
```

This issue has had a few slightly different reports collapsed into it (#603); I could only reproduce this specific one on current `master` — see "Investigation" below for the others.

## Root cause

For markdown/asciidoc output, the resources templates decide between the `No resources.` placeholder and the actual table by checking `{{- if not .Module.Resources -}}` — but `.Module.Resources` at that point is the module's **raw, unfiltered** resource list. (For `json`/`toml`/`xml`/`yaml` output, `copySections` in `format/util.go` pre-filters the module before it reaches the formatter, via `filterResourcesByMode` — but the template-based formats never call `copySections` at all; `markdown_table.go` etc. pass the raw module straight into `withModule`.)

The table body then applies a **second**, per-entry filter inside the `range` loop, based on `Sections.Resources`/`Sections.DataSources`. So a module with e.g. one managed resource and `Sections.Resources = false` looks non-empty at the top-level check (the raw list has one entry) → renders the header + table skeleton, but every row gets filtered out inside the loop → zero rows.

## Fix

Added `template.VisibleResources(resources, config)`, which applies the managed/data-source filtering once (same logic as the removed per-entry `$isResource`/`$isDataResource` checks), exposed to templates as `visibleResources`. Updated all four resources templates (`markdown_table_resources.tmpl`, `markdown_document_resources.tmpl`, `asciidoc_table_resources.tmpl`, `asciidoc_document_resources.tmpl`) to filter once up front, check emptiness against that filtered result, then range over it directly with no further per-row filtering.

## Investigation

Issue #603 has three commenters describing slightly different symptoms. I reproduced and fixed the first (mission-lblaney's, matching the exact scenario above). I could not reproduce the original reporter's exact case (`hide: [resources, data-sources]` still showing a data-source row) on current `master` — that specific combination already correctly renders nothing today, so it looks like something else fixed it since v0.16.0 (the version in that report). augustovictor's comment ("hides items of type `data` while keeping `resource` types" with `--hide data-sources`) describes correct, expected behavior rather than a bug.

## Test plan

- [x] `go build ./...`, `go vet ./...`
- [x] Added `TestVisibleResources` in `template/template_test.go`: 6 cases covering both-shown, resources-only, data-sources-only, neither-shown, the exact #603 regression scenario (module has only the hidden kind), and empty input
- [x] `go test ./...` - all green (ran in a Linux container to sidestep CRLF noise from my Windows dev checkout affecting unrelated golden-file comparisons across the whole suite; confirmed those specific failures are pre-existing/environment-only and unrelated to this change)
- [x] Manually verified all three reported scenarios end-to-end with a built binary:
  - hiding both `resources`+`data-sources` on a module with both kinds → section omitted entirely (unchanged, already correct)
  - hiding only `resources` on a module with both kinds → data source row still shows correctly (unchanged, already correct)
  - hiding `resources` on a module with *only* a managed resource → now shows `No resources.` instead of an empty table (the fix)

Fixes #603

### DIFF
--- a/format/templates/asciidoc_document_resources.tmpl
+++ b/format/templates/asciidoc_document_resources.tmpl
@@ -1,5 +1,6 @@
 {{- if or .Config.Sections.Resources .Config.Sections.DataSources -}}
-    {{- if not .Module.Resources -}}
+    {{- $resources := visibleResources .Module.Resources -}}
+    {{- if not $resources -}}
         {{- if not .Config.Settings.HideEmpty -}}
             {{- indent 0 "=" }} Resources
 
@@ -9,13 +10,9 @@
         {{- indent 0 "=" }} Resources
 
         The following resources are used by this module:
-        {{ range .Module.Resources }}
-            {{- $isResource := and $.Config.Sections.Resources ( eq "resource" (printf "%s" .GetMode)) }}
-            {{- $isDataResource := and $.Config.Sections.DataSources ( eq "data source" (printf "%s" .GetMode)) }}
-            {{- if or $isResource $isDataResource }}
-                {{- $fullspec := ternary .URL (printf "%s[%s]" .URL .Spec) .Spec }}
-                - {{ $fullspec }} {{ printf "(%s)" .GetMode -}}
-            {{- end }}
+        {{ range $resources }}
+            {{- $fullspec := ternary .URL (printf "%s[%s]" .URL .Spec) .Spec }}
+            - {{ $fullspec }} {{ printf "(%s)" .GetMode -}}
         {{- end }}
     {{ end }}
 {{ end -}}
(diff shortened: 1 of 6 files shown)


pascal-hofmann (approved):
LGTM. Thank you for the contribution.

